PR #106, opened 2015-01-22
Clarify that run_bioblend_tests.sh actively modifies the configuration o...

dannon:
...f the target instance (-g).

Subsequent commits will make this same script target a running galaxy instance without manipulating its configuration (like nosetests currently does when BIOBLEND_GALAXY_URL and BIOBLEND_GALAXY_API_KEY are set).

### DIFF
--- a/run_bioblend_tests.sh
+++ b/run_bioblend_tests.sh
@@ -4,10 +4,13 @@ show_help () {
   echo "Usage:  $0 -g GALAXY_DIR [-b BIOBLEND_DIR] [-p PORT] [-t BIOBLEND_TESTS] [-r GALAXY_REV]
 
   Run tests for BioBlend. Useful for Continuous Integration testing.
+  *Please note* that this script modifies the configuration of the galaxy
+  instance target (-g), so only use it on 'test' or otherwise disposable
+  instances.
 
 Options:
   -g GALAXY_DIR
-      Path of the local Galaxy Mercurial repository.
+      Path of the local Galaxy Mercurial repository.  The configuration of this instance will be modified to facilitate testing.
   -b BIOBLEND_DIR
       Path of the local BioBlend sources. Defaults to the current directory.
   -p PORT


jmchilton:
+1 Good clarification! Targeting an existing instance is also a good idea.

Using the newer environment variable-based configuration one can accomplish something like `-g` this without modifying Galaxy's configuration at runtime - this is what planemo does. This is probably what `-g` should actually do here.

dannon:
Totally agree, it should just set whatever environment variables are necessary.  Shutdown of the daemon also appears to be broken; investigating.

nsoranzo:
Thanks for the patch and comments, I'd also prefer the environment variable approach like planemo. If you have a patch for that, we can go directly with that approach, or I can try to write one.

dannon:
I don't have a patch ready to go or anything -- if you'd like to take a stab at implementing it go right ahead.

nsoranzo:
I'm trying to use ```GALAXY_CONFIG_FILE``` environment variable, but I just discovered that rolling_restart.sh (which I use because it waits for successful server start) treats this variable quite differently than run.sh, which is probably a bug:

rolling_restart.sh:

```
GALAXY_CONFIG_FILE=config/galaxy.ini
if [ ! -f $GALAXY_CONFIG_FILE ]; then
    GALAXY_CONFIG_FILE=universe_wsgi.ini
fi
```

run.sh:

```
if [ -z "$GALAXY_CONFIG_FILE" ]; then
    if [ -f universe_wsgi.ini ]; then
        GALAXY_CONFIG_FILE=universe_wsgi.ini
    elif [ -f config/galaxy.ini ]; then
        GALAXY_CONFIG_FILE=config/galaxy.ini
    else
        GALAXY_CONFIG_FILE=config/galaxy.ini.sample
    fi
    export GALAXY_CONFIG_FILE
fi
```

dannon:
You're right, these should behave the same way; run.sh is currently correct.

nsoranzo:
In fact, after changing rolling_restart.sh it works. I'm preparing a PR for galaxy-central.

dannon:
Sounds great!  Should we go ahead and merge this and continue the discussion of targeting (without manipulating) a running galaxy instance in an issue or elsewhere since it's a larger change?

nsoranzo:
Configuration files are no more overwritten in 869ba161582667a709bd5511311fb2beb4fb6abe . Still overwrites universe.sqlite .

Issue #108, opened 2015-02-17
Tools without input files break workflows.list()


jmeppley:
I have a workflow that includes a tool that has no input file. This tool creates a file from the input parameters only.

Line 212 in bioblend.galaxy.objects.wrapper.py:

```assert self.input_ids == set(self.inputs)```

Compares the declared inputs to the number of starting points in the DAG. Since this tool has no input files, it looks just like an input file in the DAG layout and gets put into the variable `self.input_ids`. But, it is NOT an input, so the above assertion fails.

I can solve the problem by removing the assertion, but I'm not sure what else that breaks.  The other option is to loop through the inputs and check them in the workflow data. The actual input files should have type set to 'data_input', while the tool should have type 'tool'.

nsoranzo:
@jmeppley Thanks for the detailed bug report!

Can you test if it is fixed in my tree https://github.com/nsoranzo/bioblend ?

jmeppley:
Looks good. Thanks.

PR #110, opened 2015-02-18
Rename input_ids and output_ids Workflow attributes to source_ids and sink_ids respectively

nsoranzo:
As explained in issue #108 , ```input_ids``` Workflow attribute is not properly defined: it seems to be the same of ```data_input_ids``` attribute (i.e. ids of the workflow data inputs), while it is the set of ids of the source vertices of the workflow DAG, which can be either tools or data_inputs. Similarly for ```output_ids```.

I propose to rename both ```input_ids``` and ```output_ids``` to ```source_ids``` and ```sink_ids``` respectively to avoid confusion, even if this will remove these attributes from the Workflow objects.


simleo:
Looks good and Travis agrees. Merging

PR #112, opened 2015-02-23
Restore initial 'F' in library folder IDs. Fix #111.

### LINKED ISSUE #111: Library.get_folder() incompatible with Library.folder_ids
When I try to get a folder using one of the ids found in Library.folder_ids, I get an error (or an arbitrary wrong object if the ids happen to line up).

It looks to me like the LibraryContentInfo class is stripping the leading "F" from the folder ids. I'm not sure why this is. The result is that the Galaxy API does not recognize the ID when passed to /api/libraries/LIBRARY_ID/content/FOLDER_ID

If I pull out the 'wrapped' dict from the folder's content_info object and access the u'id' value directly, I can get the Folder from Library.get_folder().

Here is the code that produces the error (library is a library object) and works around using library.content_infos:

```
library = ngi.libraries.get(libraryID)
content_info_ids=[c.wrapped['id'] for c in library.content_infos]
print "Library %s(%s)\nfolder ids: %r\ncontent info ids: %r" % (library.name,library.id, library.folder_ids, content_info_ids)

# get using folder_ids
try:
    folder=library.get_folder(library.folder_ids[0])
    print "folder_ids method: " + repr(folder.wrapped)
except Exception as e:
    # bioblend will report the error
    print "folder_ids method FAILED" 

# get using dict wrapped content_info
try:
    folder=library.get_folder(library.content_infos[0].wrapped['id'])
    print "content_infos method: " + repr(folder.wrapped)
except Exception as e:
    # bioblend will report the error
    print "content_infos method FAILED"
```

This prints:
```
Library MG ASsembly: HOT237_3_0125m(f597429621d6eb2b)
folder ids: [u'f597429621d6eb2b']
content info ids: [u'Ff597429621d6eb2b']

ERROR:bioblend:GET: error 400: '<html>\r\n  <head><title>Bad Request</title></head>\r\n  <body>\r\n    <h1>Bad Request</h1>\r\n    <p>The server could not comply with the request since\r\nit is either malformed or otherwise incorrect.\r\n\r\n<br/>Invalid LibraryDataset id ( f597429621d6eb2b ) specified\r\n<!--  --></p>\r\n    <hr noshade>\r\n    <div align="right">WSGI Server</div>\r\n  </body>\r\n</html>\r\n', 0 attempts left
folder_ids method FAILED

content_infos method: {u'parent_library_id': u'f597429621d6eb2b', u'update_time': u'2015-02-22T23:59:55.268456', u'library_path': [], u'deleted': False, u'description': u'', u'item_count': 0, u'parent_id': u'F2100007bb1a6035d', u'genome_build': None, u'model_class': u'LibraryFolder', u'id': u'Ff597429621d6eb2b', u'name': u'MG ASsembly: HOT237_3_0125m'}
```

nsoranzo:
I don't know exactly why the removing of the initial 'F' was done in the first place, so opening a PR.

### DIFF
--- a/bioblend/galaxy/objects/wrappers.py
+++ b/bioblend/galaxy/objects/wrappers.py
@@ -644,8 +644,6 @@ class LibraryContentInfo(ContentInfo):
     """
     def __init__(self, info_dict, gi=None):
         super(LibraryContentInfo, self).__init__(info_dict, gi=gi)
-        if self.id.startswith('F'):
-            object.__setattr__(self, 'id', self.id[1:])
 
     @property
     def gi_module(self):
@@ -1078,8 +1076,6 @@ class Folder(Wrapper):
 
     def __init__(self, f_dict, container_id, gi=None):
         super(Folder, self).__init__(f_dict, gi=gi)
-        if self.id.startswith('F'):  # folder id from library contents
-            object.__setattr__(self, 'id', self.id[1:])
         object.__setattr__(self, 'container_id', container_id)
 
     @property
--- a/tests/TestGalaxyObjects.py
+++ b/tests/TestGalaxyObjects.py
@@ -409,8 +409,9 @@ class TestLibrary(GalaxyObjectsTestBase):
         self.assertEqual(folder.container_id, self.lib.id)
         self.assertEqual(len(self.lib.content_infos), 2)
         self.assertEqual(len(self.lib.folder_ids), 2)
-        self.assertEqual(len(self.lib.dataset_ids), 0)
         self.assertTrue(folder.id in self.lib.folder_ids)
+        retrieved = self.lib.get_folder(folder.id)
+        self.assertEqual(folder.id, retrieved.id)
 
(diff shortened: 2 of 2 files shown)


simleo:
The reasoning behind this was that since we were building an OO API where everything was supposed to have a well-defined type, there was no need to keep a leading F on IDs.  Of course, if that's what Galaxy expects, any method that wants to talk back to the framework should re-add the F on the fly.  Or we can just merge this PR and expose those IDs to the client as they are, maybe it's not such a big deal :)

martenson:
I am afraid Galaxy expects it and there is no way around right now.
```python
if ( ( len( encoded_folder_id ) % 16 == 1 ) and encoded_folder_id.startswith( 'F' ) ):
    cut_id = encoded_folder_id[ 1: ]
else:
    exception
```

nsoranzo:
Thanks Simone and Martin for the comments!

Since also plain BioBlend does not remove the 'F' prefix, I'm merging the PR.

PR #113, opened 2015-02-23
BioBlend.objects: simple addition to get status details of an history

gmauro:


nsoranzo:
Ciao Mauro,
I like the addition of 'state' attribute, not so much the get_status() method. Would it be enough for your needs if we instead expose also the state_details dict as an attribute of History objects?

gmauro:
Ciao Nicola,

I am in the need to monitor the execution of an history and I agree with you, 'state' and 'state_details' attributes are enough.

nsoranzo:
Ok, thanks for the feedback, I will cherry pick 3619039 and f1ea474 .

Issue #114, opened 2015-02-23
Feature request: objects: Folder object attributes and methods


jmeppley:
Once you have a Folder object from a Library, there is not much in the way of operations you can do.

Minimally, the parent attribute could point back to the Library or parent Folder.

Also, it'd be useful, but harder to implement, to be able to list, retrieve, and add datasets or sub-folders using the Folder object.

nsoranzo:
Commit b408081dc5578bdc173708fc582c8a0ba777f50e adds the container attribute to Folder, which point back to the Library.

jmeppley:
Thanks.

I was mucking about in my own fork. The folder data contains a "parent_id" key that seems to point to the parent folder (if the folder is not the library's root). I added a bit of code that would set the 'parent' attribute for Folders using that. I can set up a pull request if you'd be interested, but I don't have a full testing set up to properly check it out.

nsoranzo:
> Also, it'd be useful, but harder to implement, to be able to list, retrieve, and add datasets or sub-folders using the Folder object.

All these actions should already be available through a `Library` object, I'm not sure reimplementing them inside the `Folder` class makes sense.

I'm closing this now, please open new issues for specific requests.

PR #115, opened 2015-02-25
non-root Folder automatically gets parent attribute

jmeppley:
As per discussion in #114, simply uses the 'parent_id' value in the wrapped Folder data to fetch the indicated parent Folder and set as the current Folder's 'parent' attribute.

### DIFF
--- a/bioblend/galaxy/objects/wrappers.py
+++ b/bioblend/galaxy/objects/wrappers.py
@@ -1107,8 +1107,7 @@ class Folder(Wrapper):
                 # older Galaxy versions strip the F
                 parent_id = u'F%s' % (parent_id)
             try:
-                parent = self.container.get_folder(parent_id)
-                return parent
+                return self.container.get_folder(parent_id)
             except bioblend.galaxy.client.ConnectionError:
                 # Depending on version, galaxy is returning a dummy parent_id 
                 #  for the root Folder
@@ -1116,8 +1115,7 @@ class Folder(Wrapper):
                 self.wrapped['dummy_parent_id']=self.wrapped['parent_id']
                 self.wrapped['parent_id']=None
 
-        # folder is root, return library
-        return self.container
+        return None
 
     @property
     def gi_module(self):
--- a/bioblend/galaxy/objects/wrappers.py
+++ b/bioblend/galaxy/objects/wrappers.py
@@ -1102,8 +1102,7 @@ class Folder(Wrapper):
         Return folder indicated by 'parent_id'
         """
         parent_id = self.wrapped.get('parent_id',None)
-        # 'F21000...' seems to be a dummy ID (encoding of 0 or None?)
-        if parent_id is not None and parent_id != 'F2100007bb1a6035d':
+        if parent_id is not None:
             if parent_id[0]!='F':
                 # older Galaxy versions strip the F
                 parent_id = u'F%s' % (parent_id)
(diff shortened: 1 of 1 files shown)


nsoranzo:
Hi John, that would work, but it's recursive and may create an (unwanted) chain of objects. I was thinking to use a ```@property```, so that the ```.parent``` Folder object is created on demand, and then saved in a ```_cached_parent``` attribute.

What do you think?

jmeppley:
Yes, that sounds like an improvement. I am so used to avoiding galaxy libraries with too much in them (because the web interface loads so slowly), that it never occurred to me to fix the problem on the API end. 

Unfortunately, the `.parent` `@property` is colliding with the `.parent` attribute set by the parent `class` `Wrapper`. We'd have to either change `.parent` to a `@property` in Wrapper or use another name for the `Folder` hierarchy.

jmeppley:
I've gone ahead and done the latter: changing `.parent` to a property for all `Wrapper` objects. 

I ran into a few issues: 
My older galaxy instance returns None as a parent_id of root folders, but the latest galaxy code is returning the id `F2100007bb1a6035d`. I'm assuming that's an encodeing of 0 or None, and hard-coded it into the parent property.

I also wasn't sure how to terminate the parent hierarchy. I decided to set the parent of the root folder to the library. I could have gone with None or itself.

nsoranzo:
Hi John,
thanks for working on this (I was also). The parent_id of root folders was indeed broken in Galaxy, I reported this issue 6 hours ago in the Galaxy IRC channel and it was fixed in https://github.com/galaxyproject/galaxy/commit/d3ed79bed95a6d22ff8400b1a0cb94c225534ca6 .
I will review your changes tomorrow.

nsoranzo:
Unfortunately the value ```F2100007bb1a6035d``` is not universal, but depends on the value of ```id_secret``` in config/galaxy.ini , so https://github.com/jmeppley/bioblend/commit/7ff98c8926468866d438322236db0e5a66f8dfd0#diff-caf733f4e127706515d8f72554ba05a5R1106 should be reworked.

jmeppley:
D'oh. Right. I'll take that out. What do you think about the parent of the root Folder? The Library? Itself? or None?

dannon:
I would think parent of a root folder should be None.

jmeppley:
I went back and forth a few times. I agree that it makes more logical sense. It's a little clunky because None is the default value and triggers a check of the `parent_id` in `wrapped` for `Folder` objects. 

There may be a different way to cache the parent object that gets around this, but I haven't come up with one yet.

It's not a big performance hit, so maybe it's worth just leaving like this.

nsoranzo:
I agree with Dannon, thanks John for continuing your work on this!

Still need to add unit tests to tests/TestGalaxyObjects.py and test on previous Galaxy releases. I will try to find some time tomorrow.

bioblend/galaxy/objects/wrappers.py
```diff
@@ -1079,6 +1086,38 @@ def __init__(self, f_dict, container, gi=None):
...
+                               self._get_parent())
+        return self._cached_parent
+
+    def _get_parent(self):
+        """
+        Return folder indicated by 'parent_id'
+        """
+        parent_id = self.wrapped.get('parent_id',None)
+        if parent_id is not None:
+            if parent_id[0]!='F':
```
nsoranzo: PEP8: missing spaces around "!=" operator.

bioblend/galaxy/objects/wrappers.py
```diff
@@ -1079,6 +1086,38 @@ def __init__(self, f_dict, container, gi=None):
...
+                # older Galaxy versions strip the F
+                parent_id = u'F%s' % (parent_id)
+            try:
+                return self.container.get_folder(parent_id)
+            except bioblend.galaxy.client.ConnectionError:
+                # Depending on version, galaxy is returning a dummy parent_id 
+                #  for the root Folder
+                # Clear the parent_id, so we don't try to load it each time
+                self.wrapped['dummy_parent_id']=self.wrapped['parent_id']
+                self.wrapped['parent_id']=None
```
nsoranzo: PEP8: missing spaces around "=" in the 2 previous lines.

jmeppley: Sorry, I don't do pep8 by habit yet. I have the checker installed. I'll
work through the errors.

nsoranzo:
Thanks a lot John, I merged your PR and followed up on commit cfcc4c932c21bac3062d16e28a67c75a79b0d3b1.

Issue #84, opened 2014-08-22
Use actual API endpoint for dataset downloads (at least for HDAs).


jmchilton:
http://dev.list.galaxyproject.org/Problems-downloading-files-via-bioblend-td4665435.html

erasche:
HA!

> Yes sorry about this - Eric reported this same problem this week. 

And I just re-discovered it and am bumping it again.

erasche:
Testing a fix now....

erasche:
#117

PR #117, opened 2015-02-28
Fix for #84 and documentation improvements

erasche:
Modify bioblend.galaxy.histories to implement TODOs, fix dataset URL by downloading from url provided by show_datasets() call.

This doesn't pass test cases on my desktop (but then again, reverting to the parent commit also didn't let it pass tests), so I'm going to let Travis test it and rely on those test results.

There is a small behaviour change here:

| file_path | use_default_filename | original output | new output
--- | --- | --- | ----
`str` (folder) | True | None | return path the file was saved to, rather than leaving the user guessing/calculating
`str` (file) | False| None | return the user requested path that the file was saved to
None | True or False | file contents | file contents

The other small behaviour change being that `download_dataset` in `bioblend.galaxy.histories` now returns the value of the call from `bioblend.galaxy.datasets` rather than `None`. 

Lastly, `HistoryClient.download_dataset` now ignores `history_id`, as that isn't actually needed to download a dataset.

### DIFF
--- a/bioblend/galaxy/datasets/__init__.py
+++ b/bioblend/galaxy/datasets/__init__.py
@@ -23,7 +23,13 @@ class DatasetClient(Client):
         Display information about and/or content of a dataset. This can be a
         history or a library dataset.
 
-        :type hda_ldda: string
+        :type dataset_id: str
+        :param dataset_id: Encoded Dataset ID
+
+        :type deleted: bool
+        :param deleted: Whether to return results for a deleted dataset
+
+        :type hda_ldda: str
         :param hda_ldda: Whether to show a history dataset ('hda' - the default) or library
                          dataset ('ldda').
         """
@@ -37,18 +43,18 @@ class DatasetClient(Client):
         """
         Downloads the dataset identified by 'id'.
 
-        :type dataset_id: string
+        :type dataset_id: str
         :param dataset_id: Encoded Dataset ID
 
-        :type file_path: string
+        :type file_path: str
         :param file_path: If the file_path argument is provided, the dataset will be streamed to disk
                           at that path (Should not contain filename if use_default_name=True).
                           If the file_path argument is not provided, the dataset content is loaded into memory
                           and returned by the method (Memory consumption may be heavy as the entire file
                           will be in memory).
 
-        :type use_default_name: boolean
-        :param use_default_name: If the use_default_name parameter is True, the exported
+        :type use_default_filename: boolean
+        :param use_default_filename: If the use_default_name parameter is True, the exported
                                  file will be saved as file_path/%s,
                                  where %s is the dataset name.
                                  If use_default_name is False, file_path is assumed to
@@ -137,6 +143,9 @@ class DatasetClient(Client):
     def show_stderr(self, dataset_id):
         """
         Display stderr output of a dataset.
+
+        :type dataset_id: str
+        :param dataset_id: Encoded Dataset ID
         """
         res = urllib2.urlopen(self.url[:-len("/api/datasets/")+1]+"/datasets/"+dataset_id+"/stderr")
         return res.read()
@@ -144,6 +153,9 @@ class DatasetClient(Client):
     def show_stdout(self, dataset_id):
         """
         Display stdout output of a dataset.
+
+        :type dataset_id: str
+        :param dataset_id: Encoded Dataset ID
         """
         res = urllib2.urlopen(self.url[:-len("/api/datasets/")+1]+"/datasets/"+dataset_id+"/stdout")
         return res.read()
--- a/bioblend/galaxy/datasets/__init__.py
+++ b/bioblend/galaxy/datasets/__init__.py
@@ -33,7 +33,7 @@ class DatasetClient(Client):
         return Client._get(self, id=dataset_id, deleted=deleted, params=params)
 
     def download_dataset(self, dataset_id, file_path=None, use_default_filename=True,
-                         wait_for_completion=False, maxwait=12000):
+                         wait_for_completion=False, maxwait=12000, file_ext=None):
         """
         Downloads the dataset identified by 'id'.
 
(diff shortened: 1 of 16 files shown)


bioblend/galaxy/histories/__init__.py
```diff
@@ -16,6 +13,7 @@ class HistoryClient(Client):
 
     def __init__(self, galaxy_instance):
         self.module = 'histories'
+        self.galaxy_instance = galaxy_instance
```
nsoranzo: This is not needed, ```galaxy_instance``` is already saved as ```self.gi``` in the superclass ```__init__()``` method.

erasche: Ah, excellent. Will update presently.

bioblend/galaxy/histories/__init__.py
```diff
@@ -261,26 +259,10 @@ def download_dataset(self, history_id, dataset_id, file_path,
...
-
-        req = urllib2.urlopen(url)
-        if use_default_filename:
-            file_local_path = os.path.join(file_path, meta['name'])
-        else:
-            file_local_path = file_path
-
-        with open(file_local_path, 'wb') as fp:
-            shutil.copyfileobj(req, fp)
+        dc = DatasetClient(self.galaxy_instance)
```
nsoranzo: Use ```self.gi``` instead of ```self.galaxy_instance```.

erasche: fixed

jmchilton:
:+1:

erasche:
@nsoranzo any other comments or can this be merged?

nsoranzo:
Sorry Eric, I'm working on another open PR, will get to this soon.

erasche:
Okay, no worries, just wanted to poke someone on bioblend about it. Thanks.

erasche:
Okay...I hadn't intended to include these documentation changes in this PR, but due to carelessness I did. 

Please address questions in #118 before merging this PR so I can add those improvements in.

bioblend/galaxy/datasets/__init__.py
```diff
@@ -62,6 +68,11 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
         :param maxwait: Time (in seconds) to wait for dataset to complete.
                         If the dataset state is not complete within this time, a DatasetTimeoutException will be thrown.
 
+        :type file_ext: str
+        :param file_ext: Extension to request from Galaxy. Will default to
+                         dataset file_ext value. Provided for backwards
+                         compatability with HistoryClient.download_dataset()
```
nsoranzo: s/compatability/compatibility/

bioblend/galaxy/histories/__init__.py
```diff
@@ -260,33 +337,25 @@ def download_dataset(self, history_id, dataset_id, file_path,
...
-        if use_default_filename:
-            file_local_path = os.path.join(file_path, meta['name'])
-        else:
-            file_local_path = file_path
 
-        with open(file_local_path, 'wb') as fp:
-            shutil.copyfileobj(req, fp)
+        Legacy use only, refer to
+        ``bioblend.galaxy.dataset.DatasetClient.download_dataset()`` for
+        available parameters
```
jmchilton: Hmm... structurally I love that this is no longer duplicated and is now doing the right thing. Not sure about calling this endpoint for legacy use only tough. I may prefer the idea of having two logical ways to download datasets (a history content centric way and a dataset centric way - depending on your current flow).

Thoughts?

bioblend/galaxy/workflows/__init__.py
```diff
@@ -137,21 +160,20 @@ def export_workflow_json(self, workflow_id):
 
     def export_workflow_to_local_path(self, workflow_id, file_local_path, use_default_filename=True):
         """
-        Exports a workflow in json format to a given local path.
+        Exports a workflow to a given local path.
```
nsoranzo: The file _is_ in JSON format, please revert this change.

bioblend/galaxy/workflows/__init__.py
```diff
@@ -16,15 +16,25 @@ def get_workflows(self, workflow_id=None, name=None, deleted=False, published=Fa
...
+        :param name: Filter by name of workflow (incompatible with
+          ``workflow_id``). If multiple names match the given name, all the
+          workflows matching the argument will be returned.
+
+        :type deleted: bool
+        :param deleted: If set to ``True``, return workflows that have been
+          deleted.
+
+        :type published: bool
+        :param published: If set to ``True``, return published workflows.
```
nsoranzo: There has been a merge problem here, please remove the documentation for ```deleted``` and ```published``` above.

bioblend/galaxy/workflows/__init__.py
erasche: Indeed it is

bioblend/galaxy/histories/__init__.py
erasche: Yay for less duplication.

I'm not sure I really understand the history-centric view of datasets, or rather why history was even a part of the dataset URL. I can understand wanting datasets from a history (and this convenience method for accessing it via the same class), but to my way of thinking accessing history information and accessing datasets wouldn't share the Galaxy API route.

bioblend/galaxy/workflows/__init__.py
jmchilton: Done - https://github.com/galaxyproject/bioblend/commit/beac15289a1f5b2619d72cbdd0dbc02a91de2f53. Thanks for the catch.

bioblend/galaxy/histories/__init__.py
```diff
@@ -167,16 +234,12 @@ def update_dataset(self, history_id, dataset_id, **kwds):
...
         :type history_id: str
         :param history_id: Encoded history ID
-        :type name: str
-        :param name: Replace history dataset name with the given string
-        :type annotation: str
-        :param annotation: Replace history dataset annotation with given string
-        :type deleted: bool
-        :param deleted: Mark or unmark history dataset as deleted
-        :type visible: bool
-        :param visible: Mark or unmark history dataset as visible
```
nsoranzo: Why did you remove this part of the documentation? These are the parameters that may be specified in ```**kwds```.

erasche: Hmmm. I think I removed them because they weren't explicit function parameters. That was wrong of me, I should've added them as extra arguments.

PR #119, opened 2015-03-03
API break: remove 'deleted' parameter from WorkflowClient.get_workflows() and ObjWorkflowClient.list()

nsoranzo:
Passing ```deleted=True``` to this methods has never worked since workflows are just permanently deleted.

This may break scripts calling these methods with positional arguments. Alternatively I can leave the parameter in place, but not pass it to Galaxy and deprecate it in the methods documentation.

@afgane @simleo Opinions?

### DIFF
--- a/bioblend/galaxy/objects/client.py
+++ b/bioblend/galaxy/objects/client.py
@@ -277,7 +277,8 @@ class ObjWorkflowClient(ObjClient):
             )
         return [wrappers.WorkflowPreview(_, gi=self.obj_gi) for _ in dicts]
 
-    def list(self, name=None, deleted=False, published=False):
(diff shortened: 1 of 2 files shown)


afgane:
Given Galaxy's strong desire not to break backward compatibility, perhaps deprecating the parameter for might be more appropriate with a comment that it will be removed with bioblend 0.6.

simleo:
I agree with Enis

nsoranzo:
Ok, I've pushed commit 9443bb6a13bca41d7a82bd287d3bdaf8a5d207dc instead, which deprecates and ignores the parameter.

Thanks for the feedback!

PR #120, opened 2015-03-04
matrix builds for recent galaxy releases

erasche:
Easier than setting up jenkins jobs for ALL the things.

-  Adds an [environment variable](http://docs.travis-ci.com/user/build-configuration/#The-Build-Matrix) for galaxy version, in response to discussion https://github.com/erasche/bioblend/commit/dd89790104f00a358f8ae16905364e558ffae8c9#commitcomment-10027726
- Only support relatively recent builds in order to avoid messiness with lack of user-add API route (according to @nsoranzo), and the switch from `universe_wsgi.ini` to `galaxy.ini`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,18 @@ language: python
 python:
   - "2.7"
   - "2.6"
-
+env:
+  - GALAXY_VERSION=dev
+  - GALAXY_VERSION=release_15.01
+  - GALAXY_VERSION=release_14.10
+  
 install:
   - python setup.py install
 
 before_script:
   # Install Galaxy
   - cd $HOME
-  - wget https://github.com/galaxyproject/galaxy/archive/dev.tar.gz
+  - wget https://github.com/galaxyproject/galaxy/archive/${GALAXY_VERSION}.tar.gz
   - tar xvzf dev.tar.gz | tail
   - cd galaxy-dev/
   # Copy sample files and fetch eggs
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ before_script:
   - cd $HOME
   - wget https://github.com/galaxyproject/galaxy/archive/${GALAXY_VERSION}.tar.gz
   - tar xvzf dev.tar.gz | tail
-  - cd galaxy-dev/
+  # Releases/dev branch are named differently
+  - cd galaxy-*/
   # Copy sample files and fetch eggs
   - ./scripts/common_startup.sh
   # Create a PostgreSQL database for Galaxy. The default SQLite3 database makes test fail randomly because of "database locked" error.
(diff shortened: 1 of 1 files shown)


erasche:
This PR is just meant to spark discussion. @jmchilton, comments?

jmchilton:
I like 14.10 - it is a good release to support, lots of big security fixes, it had just introduced environment variable based configuration which I think the bioblend testing should use (but is not yet). Supporting 14.10 and 14.10+1 is much easier than supporting 14.10 and 14.10-1.

Lets see what the test results look like and start annotating things with a decorator I guess?

.travis.yml
```diff
@@ -2,16 +2,21 @@ language: python
 python:
   - "2.7"
   - "2.6"
-
+env:
+  - GALAXY_VERSION=dev
```
nsoranzo: You're missing ```release_15.03```.

nsoranzo:
Cool, thanks Eric!

erasche:
The results of this are actually a bit fascinating:

- 15.01 fails per @nsoranzo's [comment about url key being too new](https://github.com/erasche/bioblend/commit/dd89790104f00a358f8ae16905364e558ffae8c9#commitcomment-10027726)
- 14.10 fails with `WebError 0.8a couldn't be downloaded automatically`, which I'd normally be inclined to blame on connectivity, but it's only present in 14.10, and on both python versions.

nsoranzo:
Failing of release_14.10 is due to something I recently reported and which was fixed by @dannon. Searching git log now.

nsoranzo:
@dannon May it be fixed by https://github.com/galaxyproject/galaxy/commit/3961d4834fe5c83fd27be176f94dbc6ec95dda9c and following commits?

dannon:
Possible, though confirmation would be good -- I think that's at least one
of the errors I was trying to fix at that time.

nsoranzo:
That should have been https://github.com/galaxyproject/galaxy/commit/dce9c1a5218fc3deb1a54b4521206bbf977af4ab and following commits. Indeed they were for the same error: https://botbot.me/freenode/galaxyproject/2015-02-12/?tz=Europe/London

jmchilton:
What do you guys think of testing python 2.6 with dev and then only testing 2.7 with all releases. I think we would largely be checking all the same things - but it should run faster (waste less energy anyway) and there would be fewer false negatives due to transient errors.

erasche:
+1 from me.

nsoranzo:
Ok for me too, are you going to commit it @jmchilton ?

Regarding false negatives, there should not be many since we are using PostgreSQL, not SQLite.

jmchilton:
@nsoranzo I probably won't get to this today - it was more like an aspirationional goal :). I have created an issue here https://github.com/galaxyproject/bioblend/issues/121.

I was more worried about false positives with respect to networking errors when setting things up - I see that frequently in other projects - but maybe the Galaxy eggs are more reliable than PyPI say.

nsoranzo:
Should we also test master branch?

erasche:
If galaxy follows the release diagram posted on trello, then master == latest release.

natefoo:
Yes, master is the latest release.

nsoranzo:
Thanks for the comments, master branch will not be added to the test matrix.

PR #123, opened 2015-03-17
Fixes #122

### LINKED ISSUE #122: Commit message not used
See [L250](https://github.com/galaxyproject/bioblend/blob/master/bioblend/toolshed/repositories/__init__.py#L250), mis-use of `dict()` vs `{}`, the commit message is stored in a key equal to its value, never being used.

I could've sworn I saw an email/IRC message about this recently, just discovered this while reading the code.

erasche:


afgane:
While in the `update_repository` method - it is missing a docstring... (@jmchilton)

jmchilton:
docstring huh? I know what a string is - but not sure what this doc thing is...

I'll do that today.

jmchilton:
Also, thanks for the fix @erasche!

PR #125, opened 2015-03-17
Add landscape.io badge

erasche:
Landscape is becoming quite popular and a number of other galaxy projects use it for monitoring code smell/issues.


nsoranzo:
@erasche It seems the last time the BioBlend code base was checked by Landscape was 30 Mar 2015, should we drop the badge or fix the configuration?

Issue #97, opened 2014-10-14
Python 3 Support


jmchilton:
Path forward:

- [x] Add dependency on [six](https://pypi.python.org/pypi/six).
- [ ] Switch all changed imports to use ``six.moves``, at this point all the files would be importable within Python 3 - which would help downstream projects like [planemo](https://github.com/jmchilton/planemo/commit/e41892bf321532f9ab1642761afaeb5db011c3e6).
- [x] Replace unmaintained `poster` dependency - ([requests-toolbelt](https://github.com/sigmavirus24/requests-toolbelt) may be a solution)
- [X] ~~Test dropping in [boto3](https://github.com/boto/boto3) in lieu of boto (or just wait for it to replace the project).~~ [Not needed](https://github.com/galaxyproject/bioblend/issues/97#issuecomment-86180237)
- [x] Update the actual BioBlend code for Python 3.
- [x] Use tox for testing.
- [ ] Setup a branch for Python 3 support using Travis to test against Python 3.4.

nsoranzo:
See commits from 9036f902bad7278087033f3fe595ecdf2a3b818c to ef48b84e2cb26092869f27c23fa13ce997d1972d for an initial work on this.

nsoranzo:
See pull request #131 .

nsoranzo:
I think we don't need boto3, [boto documentation](http://docs.pythonboto.org/) says: "Work is under way to support Python 3.3+ in the same codebase. Modules are being ported one at a time with the help of the open source community."

Both EC2 and S3 are listed as already supported for Python 3.

jmchilton:
Well that is fantastic! We just need to make poster an optional dependency or replace it - and then handle the dictionary iteration stuff. No promises, but I will try to find some time to take a crack at replacing poster tonight.

nsoranzo:
I was trying to extend TravisCI testing to Python 3.4, but this tries to run also Galaxy with Python 3, which obviously does not work: https://travis-ci.org/nsoranzo/bioblend/jobs/55978112 . Ideas?

jmchilton:
https://www.dominicrodger.com/tox-and-travis.html?

nsoranzo:
So all the python interpreters are installed on the Travis nodes as ```pythonX.Y```, independently of the python version used for that particular build. Correct?

nsoranzo:
> https://www.dominicrodger.com/tox-and-travis.html?

[...]

PR #131, opened 2015-03-25
WIP: Add support for Python >= 3.3 using six package.

nsoranzo:
TODO: dictionary iterator methods.

Also, still not possible to test on Python3 because of poster and boto packages.

### DIFF
--- a/bioblend/__init__.py
+++ b/bioblend/__init__.py
@@ -1,5 +1,6 @@
-import os
 import logging
+import os
+
 from bioblend.config import Config, BioBlendConfigLocations
 
 # Current version of the library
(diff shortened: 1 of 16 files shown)


jmchilton:
This is awesome @nsoranzo! Does the WIP mean you don't want me to merge this? Seems like an atomic commit to me - we can always follow up with fixes for dictionary iteration, boto, and poster?

simleo:
Nice work indeed! I have only one concern, and I've expressed it as PR #132 :)

nsoranzo:
@jmchilton If your review is positive, please merge! The WIP means that the 2nd point "Switch all changed imports to use six.moves, at this point all the files would be importable within Python 3 - which would help downstream projects like planemo." in #97 is not completed yet.

jmchilton:
Awesome.

PR #132, opened 2015-03-25
Do not swallow KeyboardInterrupt and SystemExit in Python 2

simleo:
In the course of the ongoing work on Python 3 support (see #131), some `except StandardError` have been replaced by `except Exception`. In Python2, however, this is a bit dangerous as it swallows up `KeyboardInterrupt` and `SystemExit`, which could lead to unexpected hangs. This PR should fix this problem (and be harmless albeit useless in Python 3).

### DIFF
--- a/bioblend/__init__.py
+++ b/bioblend/__init__.py
@@ -9,6 +9,8 @@ __version__ = '0.5.4-dev'
 try:
     import resource
     CHUNK_SIZE = resource.getpagesize()
+except (KeyboardInterrupt, SystemExit):  # for Python 2
+    raise
 except Exception:
     CHUNK_SIZE = 4096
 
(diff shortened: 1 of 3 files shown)


nsoranzo:
According to https://docs.python.org/2.6/library/exceptions.html#exceptions.KeyboardInterrupt and https://docs.python.org/2.6/library/exceptions.html#exceptions.SystemExit , each of this exception "inherits from BaseException instead of StandardError or Exception so that it is not accidentally caught by code that catches Exception. This allows the exception to properly propagate up and cause the interpreter to exit."

simleo:
Whoops! Right, that's only for Python < 2.5, which we don't support. Thanks for the review

PR #133, opened 2015-03-26
Replace poster with requests-toolbelt.

jmchilton:
poster is not Python 3 compatible but requests-toolbelt is - so this is providing yet one more step forward toward Python 3 support (#97).

requests-toolbelt is tested with requests >= 2.1.0 so I also updated that requirement in setup.py.

Additionally, f0a090e includes additional testing to verify new upload technique actual uploads the correct content.


bioblend/galaxyclient.py
```diff
@@ -7,16 +7,32 @@
...
+    def attach_file(self, path, name=None):
+        """
+        Attach a path to a request payload object.
+
+        :type path: string
+        :param path: Path to file to attach to payload.
+
+        :type name: string
+        :param name: Name to give file, if different than actual pathname.
+        """
```
nsoranzo: Missing `:rtype:` and `:return:`.

bioblend/galaxyclient.py
```diff
@@ -90,21 +106,23 @@ def make_post_request(self, url, payload, params=None, files_attached=False):
...
             payload = json.dumps(payload)
-            r = requests.post(url, data=payload, headers=self.json_headers,
-                              verify=self.verify, params=params)
-            if r.status_code == 200:
-                return r.json()
-            # @see self.body for HTTP response body
-            raise ConnectionError("Unexpected response from galaxy: %s" %
-                                  r.status_code, body=r.text)
+            headers = self.json_headers
+            params = params
```
nsoranzo: This is a no-op, can be removed.

bioblend/galaxyclient.py
```diff
@@ -7,16 +7,32 @@
...
-from six.moves.urllib.request import Request, urlopen
+from requests_toolbelt import MultipartEncoder
 import six
 
 from .galaxy.client import ConnectionError
 
 
 class GalaxyClient(object):
+
+    def attach_file(self, path, name=None):
```
nsoranzo: Not sure this function belongs to this class (does not use self), it may probably be moved to `bioblend/__init__.py`.

nsoranzo:
Awesome, I was too scared to touch this code!

Thanks a lot!!

bioblend/galaxyclient.py
jmchilton: Disagree that it should be removed but I have refactored it in ef2e903 as a compromise.

bioblend/galaxyclient.py
jmchilton: Moved in 8f155a4.

bioblend/galaxyclient.py
jmchilton: Added in 8f155a4.

jmchilton:
Great - thanks for the detailed review and merge @nsoranzo.

PR #134, opened 2015-03-26
Include toolshed documentation

erasche:


erasche:
Second commit is because: 

screenshot from 2015-03-26 15 40 27

afgane:
This is available on the master branch on RTD: http://bioblend.readthedocs.org/en/master/

Thanks!

PR #139, opened 2015-04-10
Verify SSL certificates by default.

nsoranzo:
I think we should stay on the safe side, also [requests](http://python-requests.org/) has [SSL cert verification](http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification) set to True by default.


erasche:
+1

afgane:
A sensible default. I think the reason for disabling that in the first place was because cloud instances have an option to use a self-signed cert...

PR #143, opened 2015-04-23
Adding methods to convert workflow formats

kellrott:
This is a first pass at integrating the 'simple' workflow format, using the 'yaml_to_workflow' code from https://github.com/galaxyproject/galaxy/blob/dev/test/api/yaml_to_workflow.py

A 'simple' workflow is more human readable. The test workflow in tests/data/paste_columns.ga translates to

```
name: paste_columns
steps:
- label: step_1
  type: input
- label: step_0
  type: input
- connect:
    input1: step_0#output
    input2: step_1#output
  label: step_2
  state:
    chromInfo: /home/simleo/hg/galaxy-dist/tool-data/shared/ucsc/chrom/?.len
    delimiter: T
  tool_id: Paste1
uuid: bdbdf826-08fa-43be-a633-05cd910bf98e
```

### DIFF
--- a/bioblend/galaxy/workflows/convert.py
+++ b/bioblend/galaxy/workflows/convert.py
@@ -1,0 +1,350 @@
+
+import json
+import uuid
+
+from collections import OrderedDict
+
+
+STEP_TYPE_ALIASES = {
+    'input': 'data_input',
+    'input_collection': 'data_collection_input',
+}
+
+def uniquify(name, taken):
+    if name not in taken:
+        return name
+
+    while True:
+        res = re.search(r'^(.*)_(\d+)$', name)
+        if res:
+            name = "%s_%s" % (res.group(1), int(res.group(2)) + 1)
+        else:
+            name = "%s_1" % (name)
+        if name not in taken:
+            return name
+
+def workflow_to_simple(workflow):
+    """
+    convert a galaxy formatted workflow to a 'simple'
+    format that is more human friendly to read
+    """
+
+    steps = workflow['steps']
+
+    step_map = {}
+    steps_out = []
+    for step_id, step in steps.items():
+        label = step.get("label", "step_%s" % (step_id))
+        label = uniquify(label, step_map)
+        step_map[step_id] = label
+
+    for step_id, step in steps.items():
+        new_step = {
+            'label' : step_map[step_id]
+        }
+        if step['type'] == 'data_input':
+            new_step['type'] = 'input'
+        if step['type'] == 'pause':
+            new_step['type'] = 'pause'
+        if step['type'] == 'tool':
+            new_step['tool_id'] = step['tool_id']
+            state = json.loads(step['tool_state'])
+            new_state = {}
+            for k,v in state.items():
+                if k not in ['__page__', '__rerun_remap_job_id__']:
+                    new_v = json.loads(v)
+                    if new_v is not None:
+                        new_state[k] = new_v
+            new_step['state'] = new_state
+            connections = {}
+            for connection_name, connection in step['input_connections'].items():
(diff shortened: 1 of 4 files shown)


nsoranzo:
Thanks Kyle for the nice contribution!

Unfortunately Travis is complaining about PEP-8 compliance, Python 3 and Python 2.6 compatibility and maybe something else.

Pinging @jmchilton for comments on the following issues:

I'm not sure this is the best place to add this functionality, maybe the Galaxy API would be more appropriate? Then it may be added to BioBlend, blend4j, whatever. This will make sure changes to the workflow 'ga' format are reflected in the 'simple' format. And no code duplication.

What about the "Common Workflow Language"? A converter for it also belongs to the Galaxy API in my opionion.

If instead this is deemed the right place, the copy-pasted code must be properly attributed inside the files, at least by adding a note with a link to the original.

jmchilton:
Thanks for the comments @nsoranzo 

I can follow up with fixes for non-Python 2.7 pythons and issue a PR to @kellrott's repo on that if we decide to merge it and I think requiring 15.03 or newer should fix the other problems Travis picked up.

I don't know that adding this functionality to Galaxy is viable - I like adding it here because I feel better about breaking backward compatibility as we refine the format. If it is decided that this is the wrong place to add it - I think a stand-alone project might be a better target. Certainly someday Galaxy should be able to consume a reasonable workflow format - but I am not sure we are ready yet - bioblend seems like the right combination of exposure and freedom during a period where we are trying to get the format right, gather use cases, and prove it useful.

About attribution - if merged I feel like this should be the canonical implementation and Galaxy will just start using this for tests once the bioblend egg is updated - so I wouldn't worry too much about this.

As for the "Common Workflow Language" - I think Galaxy is a long time from being able to support that and this format is better for Galaxy - it maps directly to Galaxy workflow constructs (it is a more direct mapping than even ``.ga`` in a way - given its simplicity).

nsoranzo:
I think BioBlend is almost as strict as Galaxy with respect to not breaking backward compatibility, so the format would have to be supported once released in a stable release.

If @afgane or others have opinions about which is the best place to add this code, please comment!

jmchilton:
This attitude is really a needless impediment to innovation in my opinion. It is much easier to fix a dependency on a version of bioblend than it is to fix a version of Galaxy and developers using libraries generally understand they have to fix stuff and adjust their code when they upgrade - in a way Galaxy users should not have to.

@kellrott - thanks for the contribution - I'll definitely help you make an awesome standalone library for this stuff if the consensus is this should not be included.

jmchilton:
Hey - was that really resolved @nsoranzo? bioblend doesn't have a "-1 kills the PR" policy - I feel like there needs to be an arm wrestling competition or something to resolve legitimate differences of opinions if @afgane isn't going to step in.

nsoranzo:
There's been no update to this PR to resolve the TravisCI errors, no other message from Kyle and you seemed to want to move this somewhere else... I just don't want to leave PRs/issues open forever.

afgane:
Functionally, this looks great to me given the current workflow format is not very user friendly. So, while this library would be an easy place to add it, it does explicitly extends the functionality of Galaxy - something this library hasn't done thus far. Personally, I hate when scripts that worked break due to an upgrade so am not in favor of purposefully exposing that option. If it's something to be experimented with, it could possibly be tagged as a 'lab' feature (but probably better as a small separate project).

jmchilton:
I have created a project for this here (https://github.com/jmchilton/gxformat2) and it is available on PyPI (https://pypi.python.org/pypi/gxformat2).

I will talk about this at the GCC and BOSC as well if my abstracts are accepted.

PR #146, opened 2015-06-11
add check to verify existence of the tools of the workflow

gmauro:
I have added a check to verify that every tool in the workflow exists in the Galaxy instance

workflow_broken

without this check,  workflow.is_runnable return True

### DIFF
--- a/bioblend/galaxy/objects/wrappers.py
+++ b/bioblend/galaxy/objects/wrappers.py
@@ -183,6 +183,7 @@ class Workflow(Wrapper):
     def __init__(self, wf_dict, gi=None):
         super(Workflow, self).__init__(wf_dict, gi=gi)
         missing_ids = []
+        tools_list_by_id = [t.id for t in gi.tools.list()]
         for k, v in six.iteritems(self.steps):
             # convert step ids to str for consistency with outer keys
             v['id'] = str(v['id'])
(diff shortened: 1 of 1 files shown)


bioblend/galaxy/objects/wrappers.py
```diff
@@ -183,6 +183,7 @@ class Workflow(Wrapper):
     def __init__(self, wf_dict, gi=None):
         super(Workflow, self).__init__(wf_dict, gi=gi)
         missing_ids = []
+        tools_list_by_id = [t.id for t in gi.tools.list()]
```
nsoranzo: Can you please `s/gi.tools.list()/gi.tools.get_previews()/` ?

Presently they are the same function, but in the future `list()` may be more expensive to evaluate.

nsoranzo:
:+1: 

Thanks a lot for the fix, Mauro!

bioblend/galaxy/objects/wrappers.py
nsoranzo: Also, this is breaking TraviCI testing, missing a `self.`?

```python
[t.id for t in self.gi.tools.get_previews()]
```

gmauro: No, the problems is in class TestWorkflow, that use a SAMPLE_WF_DICT without a galaxy instance.
class TestWorkflow(unittest.TestCase):

    def setUp(self):
        self.wf = wrappers.Workflow(SAMPLE_WF_DICT)

nsoranzo: Ah, of course, `self.` is not needed because this is the `__init__()` method and `gi` is also a parameter!

PR #150, opened 2015-07-05
consume APIs to search tool shed repositories and tools

martenson:


nsoranzo:
You have some trailing whitespaces that break TravisCI, can you fix them? I'm fine with a squashed commit if you like.

martenson:
@nsoranzo do you think it is ok now?

bioblend/toolshed/repositories/__init__.py
```diff
@@ -35,6 +35,72 @@ def get_repositories(self):
...
         return Client._get(self)
 
+    def search_repositories(self, q, page=1, page_size=10):
+        """
+        Search for repositories in a Galaxy Tool Shed
+
+        :type  q: str
+        :param q: query string for searching purposes
+
+        :type  page: str
```
nsoranzo: Should this be an `int`?

bioblend/toolshed/repositories/__init__.py
```diff
@@ -35,6 +35,72 @@ def get_repositories(self):
...
+        :param page: page requested
+
+        :type  page_size: str
```
nsoranzo: As above, should this be an `int`?

bioblend/toolshed/tools/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        super(ToolShedClient, self).__init__(toolshed_instance)
+
+    def search_tools(self, q, page=1, page_size=10):
+        """
+        Search for tools in a Galaxy Tool Shed
+
+        :type  q: str
+        :param q: query string for searching purposes
+
+        :type  page: str
```
nsoranzo: Should this be an `int`?

bioblend/toolshed/tools/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        :param page: page requested
+
+        :type  page_size: str
```
nsoranzo: Should this be an `int`?

bioblend/toolshed/repositories/__init__.py
martenson: It is being cast to float anyways so 1 and '1'  should both work. Do you think putting int here makes more sense? If so I will change it.

nsoranzo: Yes, please, we do the same for boolean parameters.

martenson:
I forgot about this, sorry, fixed now
ping @nsoranzo

nsoranzo:
@martenson Thanks for the fixes, but I think that the merge of master is not correct, e.g. the changes to `bioblend/toolshed/__init__.py` are lost. Maybe rebase on master?

martenson:
@nsoranzo great catch! Sorry about that :(

nsoranzo:
Thanks @martenson!

PR #160, opened 2015-09-18
Fixed to bug where extra parameters were not being passed as userdata.

nuwang:
@afgane Oops, not sure how we missed this - but none of the flavour data was getting passed across.


afgane:
I haven't used this method in a while (quite clearly) so I guess that's how...

PR #161, opened 2015-09-21
Build universal wheels

natefoo:
Since BioBlend is Python 3 compatible, its wheels can be "universal".


nsoranzo:
Thanks @natefoo!

PR #174, opened 2015-11-21
add global tempfile permissions

bgruening:
To use bioblend for [automatic testing](https://github.com/bgruening/docker-galaxy-stable/blob/travis_testing/.travis.yml) of our Galaxy Docker containers we need to share the bioblend test data with the Docker container. To make this possible we need to relax the permissions a little bit.

@afgane this fixes all failing tests for me.


afgane:
And nothing less then 777 works? I'll try it but probably not before Monday.

bgruening:
Yeah, seems that I would need 7 for others.
But I found a work around: https://github.com/bgruening/docker-galaxy-stable/pull/108
So this is not needed anymore.

PR #175, opened 2015-11-27
Add bioblend.toolshed.categories.ToolShedCategoryClient .

nsoranzo:
Rename `bioblend.toolshed.repositories.ToolShedClient` class to
`bioblend.toolshed.repositories.ToolShedRepositoryClient` .
Rename `bioblend.toolshed.tools.ToolShedClient` class to
`bioblend.toolshed.tools.ToolShedToolClient` .

Ping @afgane: you added the `bioblend.toolshed.repositories.get_categories()` method in de600e1635ac70252d2b921ee6c24f27760ae0cb, but I think it never worked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,8 +100,8 @@
 
 * Added ``copy_from_dataset()`` method to ``LibraryClient``.
 
-* Added ``create_repository()`` method to ``ToolShedClient`` (thanks to Eric
-  Rasche).
+* Added ``create_repository()`` method to ``ToolShedRepositoryClient`` (thanks
+  to Eric Rasche).
 
 * Fixed ``DatasetClient.download_dataset()`` for certain proxied Galaxy
   deployments.
--- a/bioblend/toolshed/__init__.py
+++ b/bioblend/toolshed/__init__.py
@@ -1,9 +1,8 @@
 """
 A base representation of an instance of Tool Shed
 """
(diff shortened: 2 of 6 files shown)


nsoranzo:
Ping @afgane.

afgane:
Sorry - I completely missed the first ping. This looks better now with the clear separation of components. 

Although the previous impl did work:
```
from bioblend import toolshed
ts = toolshed.ToolShedInstance(url='https://testtoolshed.g2.bx.psu.edu')
ts.repositories.get_categories()
```

nsoranzo:
@afgane Ah, you're right, that's because the implementation of `bioblend.toolshed.repositories.get_categories()` is:
```
        url = urljoin(self.url, 'categories')
        return Client._get(self, url=url)
```
where `self.url` in this case is 'https://testtoolshed.g2.bx.psu.edu/api/repositories' (without a final '/'), resulting in `url` equal to 'https://testtoolshed.g2.bx.psu.edu/api/categories' .

Should I add back `bioblend.toolshed.repositories.get_categories()` calling `bioblend.toolshed.categories.get_categories()` for compatibility?

afgane:
I'd say we just move to the new version and add a note in the CHANGELOG. The new structure is more natural and having two methods just add ambiguity. If anyone runs into the problem, it's a one line change...

nsoranzo:
Ok, thanks!

PR #176, opened 2015-12-17
Update param type for create_dataset_collection()

amaclean199:
The collection_description parameter should expect a collections.CollectionDescription instead of a str.


afgane:
Thanks.

PR #181, opened 2016-01-15
Doc fixes

amaclean199:
Noticed a couple of small errors that I thought were worth fixing.


nsoranzo:
Thanks again @amaclean199, keep the patches comin'!

PR #184, opened 2016-02-22
Modify update methods of HistoryClient to return the details instead of the status code.

nsoranzo:
Also:
- modify `GalaxyClient.make_put_request()` to return the decoded response content
- added `update()` method to `HistoryDatasetAssociation`.

Asking for a review since this slightly modifies our API.


jmchilton:
:+1: Looks like a good change to me.

afgane:
Looks good to me too. Merging.

nsoranzo:
Thank you both!

PR #188, opened 2016-07-20
Stream downloaded datasets to file

glormph:
I had to download datasets >20GB today and my processes got killed :(. So I changed the requests.get code in bioblend to accept streams and download in chunks to a file. Not much work and solved my problems. Hope it fits the project. :)

### DIFF
--- a/bioblend/galaxy/datasets/__init__.py
+++ b/bioblend/galaxy/datasets/__init__.py
@@ -107,6 +107,7 @@ class DatasetClient(Client):
         # not require a key
         stream_content = file_path is not None
         r = requests.get(url, verify=self.gi.verify, stream=stream_content)
+        r.raise_for_status()
 
         if file_path is None:
             return r.content
--- a/bioblend/galaxy/datasets/__init__.py
+++ b/bioblend/galaxy/datasets/__init__.py
@@ -12,6 +12,7 @@ from six.moves.urllib.parse import urljoin
 from six.moves.urllib.request import urlopen
 
 from bioblend.galaxy.client import Client
+import bioblend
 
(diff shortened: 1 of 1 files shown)


bioblend/galaxy/datasets/__init__.py
```diff
@@ -104,7 +104,8 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
 
         # Don't use self.gi.make_get_request as currently the download API does
         # not require a key
-        r = requests.get(url, verify=self.gi.verify)
+        stream_content = file_path is not None
+        r = requests.get(url, verify=self.gi.verify, stream=stream_content)
```
nsoranzo: I just noticed that we were missing a `r.raise_for_status()` here, can you please add it?

bioblend/galaxy/datasets/__init__.py
```diff
@@ -125,7 +126,9 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
                 file_local_path = file_path
 
             with open(file_local_path, 'wb') as fp:
-                fp.write(r.content)
+                for chunk in r.iter_content(chunk_size=1024):
```
nsoranzo: Can you please use `bioblend.CHUNK_SIZE` instead of the hardcoded 1024?

You will also need to add `import bioblend` at line 14.

nsoranzo:
Thank @glormph, this is a very nice enhancement!

glormph:
Thank you for the quick response and comments @nsoranzo !

nsoranzo:
Thanks @glormph again!

PR #189, opened 2016-07-22
Wrote a small example parameter for create_dataset_collection

glormph:
Ok so I just had to figure out how to pass the CollectionDescription parameter. Pretty easy but I still thought I might as well write an example for the docs.


afgane:
Thanks.

PR #190, opened 2016-07-30
Add install_resolver_dependencies keyword to install_repository_revision

mvdbeek:
This goes together with https://github.com/galaxyproject/galaxy/pull/2554.
On older galaxy instances install_resolver_dependencies will be silently ignored.
Is that okay or should we explicitly check the galaxy version?
ping @nsoranzo

### DIFF
--- a/bioblend/galaxy/toolshed/__init__.py
+++ b/bioblend/galaxy/toolshed/__init__.py
@@ -111,7 +111,8 @@ class ToolShedClient(Client):
 
         :type install_resolver_dependencies: bool
         :param install_resolver_dependencies: Whether or not to automatically
-                                                install resolver dependencies (e.g. conda)
+                                                install resolver dependencies (e.g. conda).
+                                                This parameter is silently ignored in Galaxy ``release_2016.04`` and earlier.
 
         :type tool_panel_section_id: str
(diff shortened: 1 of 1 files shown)


bioblend/galaxy/toolshed/__init__.py
```diff
@@ -105,6 +109,10 @@ def install_repository_revision(self, tool_shed_url, name, owner,
                                                 (see http://wiki.galaxyproject.org/DefiningRepositoryDependencies
                                                 for more details)
 
+        :type install_resolver_dependencies: bool
+        :param install_resolver_dependencies: Whether or not to automatically
+                                                install resolver dependencies (e.g. conda)
```
nsoranzo:
```
This parameter is silently ignored in Galaxy ``release_2016.04`` and earlier.
```

nsoranzo:
@mvdbeek Thanks for the PR! I think that documenting it as in my inline comment should be enough.

nsoranzo:
@mvdbeek Just to be sure before I merge: ``install_tool_dependencies``, ``install_repository_dependencies`` and ``install_resolver_dependencies`` can all be `True` at the same time?

mvdbeek:
Yes, there is no problem setting any combination of these.

nsoranzo:
Cool, thanks!

PR #196, opened 2016-08-18
Allow reloading data table

mvdbeek:


bgruening:
Ah great, I just need this one! :)

mvdbeek:
Yeah, that will make https://github.com/galaxyproject/ephemeris/blob/master/ephemeris/shed_install.py#L420 so much better :)

bioblend/galaxy/tool_data/__init__.py
```diff
@@ -44,6 +44,16 @@ def show_data_table(self, data_table_id):
         """
         return Client._get(self, id=data_table_id)
 
+    def reload_data_table(self, data_table_id):
+        """
+        Reload a data table.
+
+        :type data_table_id: str
+        :param data_table_id: ID of the data table
```
nsoranzo: Can you also add a description of the return value? I think it should be the same of the `show_data_table()` method.

nsoranzo:
Thanks @mvdbeek!

PR #199, opened 2016-09-23
Fix link to cloudman

bgruening:
This is not cloudman: http://usecloudman.org :)
Thanks for reporting @blankclemens


afgane:
Thanks to both!

PR #200, opened 2016-10-14
Change method name as same as actual behavior

manabuishii:
export_workflow_json method is not return json.

Actual case and in the document , the method returns dict not JSON.

```
        :rtype: dict
        :return: Dict representing the workflow requested
```

It's very confused...


nsoranzo:
@manabuishii Thanks for reporting the issue and for the PR! Unfortunately, more changes are needed to not break the BioBlend API and other internal code that calls this method. I've created PR #201 which should take core of that. Thanks again!

PR #201, opened 2016-10-14
Deprecate import_workflow_json and export_workflow_json in favor of import_workflow_dict and export_workflow_dict

nsoranzo:
Inspired by #200, thanks @manabuishii.


manabuishii:
Thanks.
This is a lot of better than my PR.

nsoranzo:
@manabuishii Thanks again for reporting!

PR #202, opened 2016-10-19
Added param example to show format that can use multiple params per w…

glormph:
…orkflow step

To spare people from looking at the galaxy source. Unfortunately in an old branch, but I guess it fits.
:)


nsoranzo:
Thanks @glormph, I have synced this part of the documentation with the BioBlend.objects one in d3911bfe461353d7bb432689b1d8e6f67464a555 , which should also include your fix.

PR #205, opened 2016-12-19
Drop support for Galaxy release 14.02, 14.04, 14.06 and 14.08

nsoranzo:
Galaxy has stopped applying security fixes to these branches since August 2015, so they are effectively unmaintained.

@jmchilton, @afgane: Is that OK for you?


jmchilton:
I'm happy with (:+1:) dropping support - I'm not sure I agree with the reasoning though. To clarify what you said though - we don't provide all security fixes for anything but the last stable release - only patches for explicit and obvious remote code execution attacks. There have been a number of very serious XSS scripting fixes over the years that absolutely should be applied to running Galaxies (including attacks I think that could target 16.04 for instance). I can easily imagine an attacker could target admins and exploit these XSS vulnerabilities to elevate attacks. So what "we support" is 16.10 at this point in term of security - anything further back in informal and incomplete I would say. Our API support is different and somewhat more aggressive?

nsoranzo:
> So what "we support" is 16.10 at this point in term of security - anything further back in informal and incomplete I would say. 

Is that stated somewhere? I think that should be present on the Galaxy website/wiki and clearly mentioned in the release notes of each release.

> Our API support is different and somewhat more aggressive?

I don't understand this point. Is "our" referring to Galaxy or BioBlend? And what do you mean with "aggressive"?

jmchilton:
> Is that stated somewhere? I think that should be present on the Galaxy website/wiki and clearly mentioned in the release notes of each release.

Agreed it should be mentioned, I don't think it is though. Our policies in this respect are informal as far as I am aware. I created a Galaxy issue for this - https://github.com/galaxyproject/galaxy/issues/3334.

> I don't understand this point. Is "our" referring to Galaxy or BioBlend? And what do you mean with "aggressive"?

I just meant I think that we put more thought and effort into ensuring the API is consistent with the way it worked four releases ago than ensuring the last four releases are all secure. I was referring to Galaxy not bioblend - bioblend has been exceptionally good about backward compatibility since you've taken the lead on that. I think that is great.

afgane:
Fine with me too. I'll merge it...

nsoranzo:
@jmchilton Thanks for opening the Galaxy issue and for the appreciation.

Regarding security and API stability, I mainly wanted to discourage the use of these old unmaintained Galaxy releases (and save some Travis builds), keeping compatibility with them is not a big issue right now.

PR #206, opened 2017-01-24
Add tool-centric API endpoint to install dependencies

mvdbeek:
This was added 16.10

### DIFF
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -66,6 +66,19 @@ class ToolClient(Client):
         params['trackster'] = trackster
         return self._get(params=params)
 
+    def install_dependencies(self, tool_id):
+        """
+        Install dependencies for a given tool via a resolver.
+        This works only for Conda currently.
+        This functionality is available since Galaxy release_16.10
+        and is available only to Galaxy admins.
+
+        :type tool_id: str
+        :param tool_id: id of the requested tool
+        """
+        url = "%s/tools/%s/install_dependencies" % (self.gi.url, tool_id)
+        return self._post(payload={}, url=url)
+
     def show_tool(self, tool_id, io_details=False, link_details=False):
         """
         Get details of a given tool.
--- a/tests/TestGalaxyTools.py
+++ b/tests/TestGalaxyTools.py
@@ -126,3 +126,8 @@ class TestGalaxyTools(GalaxyTestBase.GalaxyTestBase):
         self.assertEqual(len(tool_output["outputs"]), 1)
         # TODO: Wait for results and verify it has 3 lines - 1 2 3, 4 5 6,
         # and 7 8 9.
+
+    @test_util.skip_unless_galaxy('release_16.10')
+    def test_tool_dependency_install(self):
+        installed_dependencies = self.gi.tools.install_dependencies('__SET_METADATA__')
+        self.assertTrue(any([True for d in installed_dependencies if d.get('name') == 'samtools' and d.get('dependency_type') == 'conda']))
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,8 @@ before_script:
   - sed -i -e 's/^#allow_user_dataset_purge.*/allow_user_dataset_purge = True/' $GALAXY_CONFIG_FILE
   # Change Galaxy configuration needed by some library tests
   - sed -i -e 's/^#allow_library_path_paste.*/allow_library_path_paste = True/' $GALAXY_CONFIG_FILE
+  # Change Galaxy configuration needed by some tool tests
+  - sed -i -e 's/^#conda_auto_init.*/conda_auto_init = True/' $GALAXY_CONFIG_FILE
   # Change Galaxy configuration needed by some workflow tests
   - sed -i -e 's/^#enable_beta_workflow_modules.*/enable_beta_workflow_modules = True/' $GALAXY_CONFIG_FILE
   # Change Galaxy configuration needed by some user tests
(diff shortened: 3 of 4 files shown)


afgane:
Can (or should) a test be written for this?
You can isolate it so it would run only on Galaxy 16.10 with this decorator `@test_util.skip_unless_galaxy('release_16.10')`

bioblend/galaxy/tools/__init__.py
```diff
@@ -66,6 +66,19 @@ def _raw_get_tool(self, in_panel=None, trackster=None):
         params['trackster'] = trackster
         return self._get(params=params)
 
+    def install_dependencies(self, tool_id):
+        """
+        Install dependencies for a given tool via a resolver.
+        This works for conda only currently.
+        This functionality is available since galaxy release 16.10
+        and is limited to admins.
```
nsoranzo: `and is available only to Galaxy admins.`

bioblend/galaxy/tools/__init__.py
```diff
@@ -66,6 +66,19 @@ def _raw_get_tool(self, in_panel=None, trackster=None):
...
+
+        :type tool_id: str
+        :param tool_id: id of the requested tool
+        """
+        url = "%s/tools/%s/install_dependencies" % (self.gi.url, tool_id)
+        return Client._post(self, url=url, payload={})
```
nsoranzo:
```python
return self._post(payload={}, url=url)
```

mvdbeek:
> Can (or should) a test be written for this?
> You can isolate it so it would run only on Galaxy 16.10 with this decorator @test_util.skip_unless_galaxy('release_16.10')

The API itself is being tested in test/integration/test_resolvers.py, which sets up a conda environment. Can we do this in bioblend?

bioblend/galaxy/tools/__init__.py
```diff
@@ -66,6 +66,19 @@ def _raw_get_tool(self, in_panel=None, trackster=None):
...
+        Install dependencies for a given tool via a resolver.
+        This works only for Conda currently.
+        This functionality is available since Galaxy release_16.10
+        and is available only to Galaxy admins.
+
+        :type tool_id: str
+        :param tool_id: id of the requested tool
+        """
+        url = "%s/tools/%s/install_dependencies" % (self.gi.url, tool_id)
+        return Client._post(self, payload={}, url=url)
```
nsoranzo: You missed one requested change on this line:
```
s/Client._post(self, /self._post(/
```

mvdbeek: oops, I just amended that.

nsoranzo:
> The API itself is being tested in test/integration/test_resolvers.py, which sets up a conda environment. Can we do this in bioblend?

Is it enough to add `conda_auto_init = True` to the Galaxy config?

mvdbeek:
I think so. Maybe we just test starting from 17.01, where this is default?

nsoranzo:
You can insert:
```bash
# Change Galaxy configuration needed by some tool tests
sed -i -e 's/^#conda_auto_init.*/conda_auto_init = True/' $GALAXY_CONFIG_FILE
```
at line 95 of `run_bioblend_tests.sh` .
I think it should be fine to decorate the forthcoming test with `@test_util.skip_unless_galaxy('release_16.10')` .

bioblend/galaxy/tools/__init__.py
```diff
@@ -66,6 +66,19 @@ def _raw_get_tool(self, in_panel=None, trackster=None):
...
+        Install dependencies for a given tool via a resolver.
+        This works only for Conda currently.
+        This functionality is available since Galaxy release_16.10
+        and is available only to Galaxy admins.
+
+        :type tool_id: str
+        :param tool_id: id of the requested tool
+        """
+        url = "%s/tools/%s/install_dependencies" % (self.gi.url, tool_id)
+        return self._post(self, payload={}, url=url)
```
nsoranzo: Remove the second `self`, this should be:
```python
return self._post(payload={}, url=url)
```

tests/TestGalaxyTools.py
```diff
@@ -126,3 +126,8 @@ def test_run_cat1(self):
         self.assertEqual(len(tool_output["outputs"]), 1)
         # TODO: Wait for results and verify it has 3 lines - 1 2 3, 4 5 6,
         # and 7 8 9.
+
+    @test_util.skip_unless_galaxy('release_16.10')
+    def test_tool_dependency_install(self):
+        installed_dependencies = self.gi.tools.install_dependencies('__SET_METADATA__')
+        self.assertTrue(any([True for d in installed_dependencies if d.get('name', None) == 'samtools' and d.get('dependency_type', None) == 'conda']))
```
nsoranzo: You can simplify this as:
```python
self.assertTrue(any(True for d in installed_dependencies if d.get('name') == 'samtools' and d.get('dependency_type') == 'conda'))
```

nsoranzo:
The new test now passes on `release_17.01`, but not on `release_16.10`.

mvdbeek:
Hmm, any way to see if the conda installation succeeded?

mvdbeek:
`requests.packages.urllib3.connectionpool: DEBUG: http://localhost:8080 "POST /api/tools/__SET_METADATA__/install_dependencies?key=21b2ff441f867dc474fb89d310d82b03 HTTP/1.1" 200 None` ... so at least the API itself seems to be OK.

nsoranzo:
Has the output of the API call changed maybe?

mvdbeek:
Slightly, yes, but what we're testing hasn't changed (first call is 16.10):
```
In [3]: gi.tools.install_dependencies('__SET_METADATA__')
Out[3]:
[{u'cacheable': True,
  u'dependency_type': u'conda',
  u'environment_path': None,
  u'exact': True,
  u'model_class': u'CondaDependency',
  u'name': u'samtools',
  u'version': None}]

In [4]: gi.tools.install_dependencies('__SET_METADATA__')
Out[4]:
[{u'cacheable': False,
  u'dependency_type': u'conda',
  u'environment_path': u'/Users/mvandenb/src/galaxy/database/dependencies/_conda/envs/__samtools@_uv_',
  u'exact': True,
  u'model_class': u'MergedCondaDependency',
  u'name': u'samtools',
  u'version': None}]
```

nsoranzo:
Ah, I know why: the change to `run_bioblend_tests.sh` needs to be applied to `.travis.yml` too! Need to integrate these 2 files better.

mvdbeek:
Alright, let's see!

nsoranzo:
Thanks a lot @mvdbeek!

Issue #208, opened 2017-01-31
Move bioblend tests into the bioblend module.


jmchilton:
If one does a pip install bioblend - the tests get installed into a tests package currently - which seems to be bad practice. They should be excluded when doing ``find_packages`` or better yet - moved and installed within the bioblend namespace.

@nsoranzo Any objections to me making the tests more usable as part of the install this way?

nsoranzo:
@jmchilton Thanks for reporting! I'd probably prefer to exclude them. @afgane ?

jmchilton:
But they are so useful :cry: - we need to work toward automated tests of Galaxy configurations and deployments and formalizing a bioblend-based test suite would be big step toward that. Also more use of that testing framework would probably result in more tests - as people find regressions, etc... they want to ensure do not happen.

nsoranzo:
I see your point, I just don't want them to be somehow considered part of the API. What's the problem with launching BioBlend tests with `./run_bioblend_tests.sh` for your use cases?

jmchilton:
I'll open a PR to be specific in answering this.

guerler:
I have issues running tests locally. I cloned a clean bioblend and galaxy repo and then called ./run_bioblend_tests.sh, the initial run somtimes works but if I run it again the test environment fails to setup properly. It complains about duplicate test users.

nsoranzo:
@guerler `./run_bioblend_tests.sh` always initialises a new SQLite database in a temporary directory, maybe the previous Galaxy instance is still around and using the port? You can also check the `main.log` inside the Galaxy repo dir.

guerler:
Its not the other Galaxy instance, when running `./run_bioblend_tests.sh` the first time it seems to process it but I cannot find details on which individual unit tests failed (is there a log file?). Instead the screen displays a long list of parsing warnings regarding missing spaces (is there a way to disable these?). When running it again, I see a warning claiming that the user already exists.

nsoranzo:
If the warnings are about stuff in `.eggs`, try to update BioBlend from GitHub, I have already fixed this in commit 3ec45d381616b29edb8e421bb94f52f6893d0c0e . Can you gist the full output of `./run_bioblend_tests.sh` ? Also, is under Linux or Mac?

guerler:
I am running it on Mac, and its the latest BioBlend repo. Here is a link to one of the outputs:
https://gist.github.com/guerler/f76e7fc9ffcd46a21bab03d4ea5cc7b4

[...]

PR #209, opened 2017-02-01
Bundle tests with bioblend including wrapper script to run them quickly.

jmchilton:
Continuing from discussion with @nsoranzo on #208.

I made a change to the original idea and placed the tests in ``bioblend/_tests`` instead of ``bioblend/test`` to make sure it is clear the tests aren't considered part of the API per se.

The advantage of this versus run_bioblend_tests is that you don't need to use bioblend from source anymore to run the tests. Simply ``pip install bioblend[testing] && bioblend-galaxy-tests`` to run the tests. This will really cleanup something like https://github.com/galaxyproject/galaxy-flavor-testing/blob/master/Makefile#L45 IMO.


jmchilton:
As long as ``_tests`` is in the Python module that is installed - I can move the driver script out into a new project - so we wouldn't need to make the changes to setup.py if they are considered outside the scope of what you'd like to do with bioblend.

setup.py
```diff
@@ -30,6 +35,10 @@
       install_requires=['requests>=2.4.3,!=2.12.0,!=2.12.1', 'requests-toolbelt', 'boto>=2.9.7', 'pyyaml', 'six'],
       tests_require=tests_require,
       packages=find_packages(),
+      entry_points=ENTRY_POINTS,
+      extras_require = {
```
nsoranzo: E251 unexpected spaces around keyword / parameter equals

setup.py
```diff
@@ -30,6 +35,10 @@
...
+        'testing':  ["pytest"],
```
nsoranzo: E121 continuation line under-indented for hanging indent
E241 multiple spaces after ':'

bioblend/_tests/pytest_galaxy_test_wrapper.py
```diff
@@ -0,0 +1,42 @@
...
+try:
+    import pytest
+except ImportError:
+    pytest = None
+
+DIRECTORY = os.path.abspath(os.path.dirname(__file__))
+BIOBLEND_TEST_SUITE = os.environ.get("BIOBLEND_TEST_SUITE", "full")
+
+quick_tests = [
+    "TestGalaxyRoles.py",
```
nsoranzo: W291 trailing whitespace

bioblend/_tests/pytest_galaxy_test_wrapper.py
```diff
@@ -0,0 +1,42 @@
...
+                args.append(os.path.join(DIRECTORY, quick_test))
+    _pytest_main(args)
+
+
+def _pytest_main(args):
+    if pytest is None:
+        raise Exception("pytest is required to use this script.")
+    print args
+    pytest.main(args)
+
```
nsoranzo: W293 blank line contains whitespace

bioblend/_tests/pytest_galaxy_test_wrapper.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+import os
+import sys
+
+try:
+    import pytest
```
nsoranzo: `setup.py` already has `nose>=1.3.1` in `tests-require`, any specific reason not to use that instead?

bioblend/_tests/pytest_galaxy_test_wrapper.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
```
nsoranzo: Can you add a short description of what this wrapper is supposed to do?

jmchilton:
@nsoranzo Before I respond to the small comments and fix things up - are you okay with this in a broad sense?

nsoranzo:
My general thoughts are about testing improvements:
- I'm fine with moving the tests to `bioblend/_tests/` with the symlink
- user creation and API key retrieval of `run_bioblend_tests.sh` should not use BioBlend itself but the Galaxy API
- the Galaxy configuration part of `run_bioblend_tests.sh` need to be split up to be reused by `.travis.yml`

With the last 2 points implemented, I suppose your wrapper can be integrated in BioBlend testing itself. In the mean time I see the benefits for testing the Galaxy API.

So I don't think I have anything against this :)

bioblend/_tests/pytest_galaxy_test_wrapper.py
jmchilton: So I feel like all the Galaxy projects should be slowly transitioning to pytest - it is more modern and better supported Python testing framework. Compare https://github.com/pytest-dev/pytest and https://github.com/nose-devs/nose for instance - it seems like pytest is more actively developed. Since this would expose the underlying testing options - I thought it would be better than nose.

I'd be happy to switch the rest of the bioblend testing stack to use pytest - it should be done anyway. I need to do the same for pulsar, planemo, and ultimately Galaxy.

jmchilton:
I removed the WIP because Travis is green again, but I'm still waiting to hear back on the pytest point.

bioblend/_tests/pytest_galaxy_test_wrapper.py
nsoranzo: Seems a good plan, would you prefer to get this merged and do the pytest migration in a new PR?

jmchilton: No - I'll bake pytest stuff into this PR.

nsoranzo: OK, I have some changes to the tests, but I can wait.

jmchilton: I don't want to hold you up - definitely feel free to merge before I get that done. I'll create an issue and assign myself.

bgruening:
ping @jmchilton @nsoranzo I also need to update the galaxy-docker testing and would like to use this is possible.

PR #210, opened 2017-02-02
Update import_shared_workflow

DamCorreia:
/api/workflows/import is deprecated since Galaxy release_14.10.
Update this method to use the new way to import shared workflow

### DIFF
--- a/bioblend/galaxy/workflows/__init__.py
+++ b/bioblend/galaxy/workflows/__init__.py
@@ -127,8 +127,8 @@ class WorkflowClient(Client):
         """
         Imports a new workflow from the shared published workflows.
 
-        :type shared_workflow_id: str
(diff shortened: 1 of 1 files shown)


bioblend/galaxy/workflows/__init__.py
```diff
@@ -123,12 +123,12 @@ def import_workflow_from_local_path(self, file_local_path):
 
         return self.import_workflow_json(workflow_json)
 
-    def import_shared_workflow(self, workflow_id):
+    def import_shared_workflow(self, shared_workflow_id):
```
nsoranzo: Can you revert the parameter name change here? This may potentially break downstream code using this as a keyword argument.

nsoranzo:
Thanks again @DamCorreia!

PR #219, opened 2017-04-04
Allow setting $BIOBLEND_GALAXY_MASTER_API_KEY instead of $BIOBLEND_GALAXY_API_KEY for tests

jmchilton:
Auto-create a user and API key in the test framework itself if this is set. Eliminates redundancies between run_bioblend_tests.sh and .travis.yml while also simplifying my external usage of bioblend-galaxy-tests.


bioblend/_tests/test_util.py
```diff
@@ -31,9 +36,40 @@ def skip_unless_galaxy(min_release=None):
...
+
+    if 'BIOBLEND_GALAXY_URL' not in os.environ:
+        return unittest.skip(NO_GALAXY_MESSAGE)
+
+    if 'BIOBLEND_GALAXY_API_KEY' not in os.environ and 'BIOBLEND_GALAXY_MASTER_API_KEY' in os.environ:
+        galaxy_url = os.environ['BIOBLEND_GALAXY_URL']
+        galaxy_master_api_key = os.environ['BIOBLEND_GALAXY_MASTER_API_KEY']
+        gi = bioblend.galaxy.GalaxyInstance(galaxy_url, galaxy_master_api_key)
+
+        if 'BIOBLEND_GALAXY_USER' in os.environ:
```
nsoranzo: I think it would be better to check for `GALAXY_USER_EMAIL` here (and skip otherwise), because that needs to be mentioned in the `admin_users` line of `config/galaxy.ini` for some tests.

.travis.yml
```diff
@@ -46,9 +46,10 @@ before_script:
   # Change Galaxy configuration to use the newly created PostgreSQL database
   - sed -e 's|^#database_connection.*|database_connection = postgresql://postgres:@localhost/galaxy-travis|' $GALAXY_CONFIG_FILE.sample > $GALAXY_CONFIG_FILE
   # Setup Galaxy master API key and admin user
-  - GALAXY_MASTER_API_KEY=$(cat /dev/urandom | LC_ALL=C tr -dc A-Za-z0-9 | head -c 32)
-  - GALAXY_USER_EMAIL=${USER}@localhost.localdomain
-  - sed -i.bak -e "s/^#master_api_key.*/master_api_key = $GALAXY_MASTER_API_KEY/" -e "s/^#admin_users.*/admin_users = $GALAXY_USER_EMAIL/" $GALAXY_CONFIG_FILE
+  - export BIOBLEND_GALAXY_MASTER_API_KEY=$(cat /dev/urandom | LC_ALL=C tr -dc A-Za-z0-9 | head -c 32)
+  - export BIOBLEND_GALAXY_USER=${USER}
+  - GALAXY_USER_EMAIL=${BIOBLEND_GALAXY_USER}@localhost.localdomain
```
nsoranzo: I think it would be better to export `GALAXY_USER_EMAIL` and drop `BIOBLEND_GALAXY_USER`, see below.

bioblend/_tests/test_util.py
```diff
@@ -31,9 +36,40 @@ def skip_unless_galaxy(min_release=None):
...
+            galaxy_user = os.environ['BIOBLEND_GALAXY_USER']
+        else:
+            galaxy_user = ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(5))
+
+        galaxy_user_id = None
+        for user in gi.users.get_users():
+            if user["username"] == galaxy_user:
```
nsoranzo: So here you would check for `galaxy_email` and get username and id, if found.

bioblend/_tests/test_util.py
```diff
@@ -31,9 +36,40 @@ def skip_unless_galaxy(min_release=None):
...
+                galaxy_user_id = user["id"]
+                break
+
+        if galaxy_user_id is None:
+            galaxy_user_email = galaxy_user + "@localhost.localdomain"
```
nsoranzo: And here you would generate a random username

.travis.yml
jmchilton: Should it be called ``BIOBLEND_GALAXY_EMAIL`` for consistency though?

jmchilton:
How about we cut the baby in half - I switch the Travis and ``run_`` tests to use a new ``BIOBLEND_GALAXY_USER_EMAIL`` but I keep the option to specify just ``BIOBLEND_GALAXY_USER``. I feel like either should work - there isn't a good reason to restrict that - it is true that some tests require an ``admin_user`` be setup - but ideally they would ``skip`` if they are not an admin.

.travis.yml
nsoranzo: `BIOBLEND_GALAXY_USER_EMAIL` looks good.

nsoranzo:
@jmchilton I see what you mean, but I think the admin tests are important and adding `BIOBLEND_GALAXY_USER` would introduce more issues/confusions (e.g. which takes precedence if both are defined?) than benefits.

tox.ini
```diff
@@ -16,4 +16,4 @@ deps =
     py{27,33,34,35}: flake8-import-order>=0.9
     nose>=1.3.1
     py26: unittest2>=0.5.1
-passenv = BIOBLEND_GALAXY_URL BIOBLEND_GALAXY_API_KEY GALAXY_VERSION
+passenv = BIOBLEND_GALAXY_URL BIOBLEND_GALAXY_API_KEY BIOBLEND_GALAXY_MASTER_API_KEY BIOBLEND_GALAXY_USER_EMAIL GALAXY_VERSION
```
nsoranzo: You can remove `BIOBLEND_GALAXY_API_KEY` now.

jmchilton: But test_utils still uses that if it is set - the master API key thing is just a fallback.

jmchilton:
Thanks for the merge @nsoranzo !

PR #225, opened 2017-05-03
More API Coverage

erasche:
As I'm playing more with parsec, I'm finding some other features not well covered so here's a PR with them. It is not a coherent collection of stuff, my apologies.

- get_users now supports server side filtering that's available
- there's a function to get a user's API key without resetting it.
- /api/version

### DIFF
--- a/bioblend/galaxy/users/__init__.py
+++ b/bioblend/galaxy/users/__init__.py
@@ -17,9 +17,10 @@ class UserClient(Client):
         Get a list of all registered users. If ``deleted`` is set to ``True``,
         get a list of deleted users.
 
-        The f_email, f_name, and f_any parameters will work for non-admin users
-        only if the Galaxy instance sets expose_user_email and expose_user_name.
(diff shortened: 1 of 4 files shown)


bgruening:
More!!! :)

martenson:
And faster! 💯

bioblend/galaxy/users/__init__.py
```diff
@@ -145,3 +167,21 @@ def delete_user(self, user_id, purge=False):
...
+        :param user_id: encoded user ID
+
+        :rtype: str
+        :return: the API key for the user
+        """
+
+        url = self.gi._make_url(self, None)
+        url = '/'.join([url, user_id, 'api_key', 'inputs'])
+        payload = {}
+        payload['user_id'] = user_id
```
nsoranzo: @erasche I don't think the `payload` is needed here, I'm going to remove it. Let me know if I'm missing something.

bgruening:
I restarted the tests, now everything passes.

bgruening:
upps, this was sended way to late ...

bioblend/galaxy/users/__init__.py
erasche: you're right, that shouldn't be needed.

erasche:
@nsoranzo thanks for tracking down and documenting all of the specific versions where these features were added. I'll try and be better about that going forward.

Issue #218, opened 2017-04-03
adding/deleting a quota


bernt-matthias:
It would be nice to have the possibility to add/edit quotas via bioblend. I have seen that the API supports this. 

Actually I could imagine to try implementing this on my own, but the API documentation on this subject is quite sparse (parameters are undocumented). Maybe I'm able to dig in the API sources to get the infos on necessary parameters. But I guess I could need some input from experienced developers.

pcm32:
I would be really interested as well in being able to set user/group quotas through bioblend. Have you had any chance to make it work @bernt-matthias?

Thanks!
Pablo

bernt-matthias:
Unfortunatelly .. no.

nsoranzo:
xref. #226

bernt-matthias:
I guess we can close this issue.

nsoranzo:
Fixed by #226.

PR #226, opened 2017-05-08
Quota API

erasche:
I didn't write any tests for adding / removing roles to quotas since that seemed like a less than easy operation to test, judging by some of the other test cases. I can go back and do this if necessary.

It seems to work fine though.

### DIFF
--- a/bioblend/_tests/TestGalaxyQuotas.py
+++ b/bioblend/_tests/TestGalaxyQuotas.py
@@ -1,0 +1,63 @@
+"""
+Use ``nose`` to run these unit tests.
+"""
+import os
+import shutil
+import tempfile
+import time
+
+from . import GalaxyTestBase, test_util
+
+
+@test_util.skip_unless_galaxy()
+class TestGalaxyQuotas(GalaxyTestBase.GalaxyTestBase):
+
+    def setUp(self):
+        super(TestGalaxyQuotas, self).setUp()
+        # Quota names must be unique, and they're impossible to delete
+        # without accessing the database.
+        self.quota_name = 'BioBlend-Test-Quota-%s' % time.time()
+        self.quota = self.gi.quotas.create_quota(
+            self.quota_name, 'testing', '100 GB', '=',
+            default='registered'
+        )
+
+    def tearDown(self):
+        self.gi.quotas.update_quota(self.quota['id'], default='registered')
+        self.gi.quotas.update_quota(self.quota['id'], default='no')
+        self.gi.quotas.delete_quota(self.quota['id'])
+        # self.gi.libraries.delete_library(self.library['id'])
+
+    def test_create_quota(self):
+        quota = self.gi.quotas.show_quota(self.quota['id'])
+        self.assertEqual(quota['name'], self.quota_name)
+        self.assertEqual(quota['bytes'], 107374182400)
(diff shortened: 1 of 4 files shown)


nsoranzo:
This should fix #218.

bernt-matthias:
Great. Exactly what I needed. Thanks.

erasche:
Will get this updated / passing shortly, apologies for the delays.

nsoranzo:
NP, I think it's just a matter of modifying `.travis.yml` and `run_bioblend_test.sh`

bioblend/_tests/TestGalaxyQuotas.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@test_util.skip_unless_galaxy()
+class TestGalaxyQuotas(GalaxyTestBase.GalaxyTestBase):
+
+    def setUp(self):
+        super(TestGalaxyQuotas, self).setUp()
+        # Quota names must be unique, and they're impossible to delete
+        # without accessing the database.
+        self.quota_name = 'BioBlend-Test-Quota-%s' % time.time()
```
nsoranzo: In `bioblend/_tests/TestGalaxyGroups.py` we use `uuid.uuid4().hex` instead of `time.time()`, I suppose it's better for random names.

erasche:
Comments addressed. Thanks for all the suggestions @nsoranzo! :)

bioblend/_tests/TestGalaxyQuotas.py
```diff
@@ -0,0 +1,59 @@
...
+        # Quota names must be unique, and they're impossible to delete
+        # without accessing the database.
+        self.quota_name = 'BioBlend-Test-Quota-%s' % uuid.uuid4().hex
+        self.quota = self.gi.quotas.create_quota(
+            self.quota_name, 'testing', '100 GB', '=',
+            default='registered'
+        )
+
+    def tearDown(self):
+        self.gi.quotas.update_quota(self.quota['id'], default='registered')
```
nsoranzo: Is this really needed, since in the next line it's updated again with `default='no'` ?

erasche: yes, unfortunately so. If `default!="no"`, then you can make any number of update_quota calls that you want. However, if `default=no` and you make a second update_quota call with default=no, then it will return an HTML error page. Documented in the [docstring](https://github.com/galaxyproject/bioblend/pull/226/files#diff-df5232b41baccd7cd9c8aa59f76b497fR139)

erasche: This API desperately needs to be rewritten using the more modern galaxy conventions. I cannot imagine anyone scripting around this API and enjoying it.

nsoranzo: Uh, awful! What about changing the method docstring from:

> Calling update twice with a default of ``no`` will throw an error.

to: 

> Calling this method with ``default="no"`` on a non-default quota will throw an error.

? I think this wording includes also the case where the quota is directly created with ``default="no"`` and then updated.

erasche: Sure, that's definitely clearer!

nsoranzo:
Thanks a lot @erasche!

Issue #221, opened 2017-05-02
 bioblend.galaxy.datasets.DatasetClient.show_stdout broken for private histories/datasets


edwardsnj:
Perhaps this represents a gap in the Galaxy API (since the show_stdout method appears to work around it), but the current strategy (see below) doesn't populate the galaxysession authentication Cookie before accessing the galaxy (non-API) url using urlopen. 

I get a HTML formatted page back (which includes text to indicate I should login).

Similarly for bioblend.galaxy.datasets.DatasetClient.show_stderr.

```
    def show_stdout(self, dataset_id):
        """
        Get the stdout output of a dataset.

        :type dataset_id: str
        :param dataset_id: Encoded dataset ID
        """
        res = urlopen(self.url[:-len("/api/datasets/") + 1] + "/datasets/" + dataset_id + "/stdout")
        return res.read()

```

nsoranzo:
@edwardsnj I think the proper way to access stdout and stderr is through the job API using `full_details=True`, see https://bioblend.readthedocs.io/en/latest/api_docs/galaxy/all.html#bioblend.galaxy.jobs.JobsClient.show_job . Feel free to reopen if this does not work for you.

edwardsnj:
Ah. Good lead on getting to the right answer - in fact it is simpler than jobs.show_job - bioblend.galaxy.histories.HistoryClient.show_dataset_provenance returns a dictionary with stdout and stderr keys (populated too! :-) I found this because I needed a way to associate jobs with histories to try out your suggestion.

Thanks!

PR #227, opened 2017-05-10
Deprecate show_stderr() and show_stdout() methods of DatasetClient

nsoranzo:
xref. https://github.com/galaxyproject/bioblend/issues/221 and https://github.com/galaxyproject/galaxy/issues/4014


nsoranzo:
Thanks @erasche, good idea! I've rebased the PR with your suggestion.

erasche:
Awesome, thanks! This looks great. :+1: (And I should be able to detect that in parsec wrappers / etc and push people toward the right calls instead. This is really awesome!)

PR #234, opened 2017-06-21
JSON header needed for delete too

erasche:
Tried to delete a history with parsec, my server threw an error.


nsoranzo:
Thanks @erasche!

PR #238, opened 2017-06-24
Include test data with install and set proper exit code

### LINKED ISSUE #237: Including test data with bioblend install
I was trying to setup the new `bioblend-galaxy-tests` script for https://github.com/ARTbio/GalaxyKickStart, but running the full test suite failed because https://github.com/galaxyproject/bioblend/tree/master/bioblend/_tests/data is not included when installing bioblend (`pip install bioblend`).

mvdbeek:
That should fix #237 .
Also propagates the exit code returned by pytest.main.


setup.py
```diff
@@ -34,7 +34,8 @@
       url="http://bioblend.readthedocs.org/",
       install_requires=['requests>=2.4.3,!=2.12.0,!=2.12.1', 'requests-toolbelt', 'boto>=2.9.7', 'pyyaml', 'six'],
       tests_require=tests_require,
-      packages=find_packages(exclude=['tests']),
+      packages=find_packages(),
```
nsoranzo: Is this first change actually needed? `tests/` directory is needed only for `run_bioblend_tests.sh`, which is supposed to be a script for developers.

mvdbeek: You're right, this isn't needed.

nsoranzo:
Thanks @mvdbeek!

PR #244, opened 2017-08-04
histories.delete_dataset dataset_id -> hda_id

lecorguille:
I'm not sure at 100% but I believe that `dataset_id` is confusing because in this case, you delete [and purge] the `history_dataset_association` not the dataset directly.

And eventually, you delete and purge the dataset if there isn't any other "association".
At least, it is what I have discovered :)


nsoranzo:
In Galaxy/BioBlend user-level documentation we rarely use the full "history dataset association" name, which is a bit low-level database-oriented terminology. What about leaving the parameter name as it was and describe it as `Encoded history dataset ID` ?

PR #248, opened 2017-09-06
Allow downloading datasets even if Galaxy reports an empty file_ext.

jmchilton:


nsoranzo:
The method already resorts to `data_type`, isn't that enough?

jmchilton:
It resorts to data_type if file_ext is not present. In a modern Galaxy I believe file_ext can be present and just be empty (""), this was needed in my application to download those files.

nsoranzo:
Galaxy bug then? Why should `file_ext` be empty?

jmchilton:
There are parts of the Galaxy code that seem adapted to handle a ``None`` extension

```
    @property
    def datatype(self):
        extension = self.extension
        if not extension or extension == 'auto' or extension == '_sniff_':
            extension = 'data'
        ret = _get_datatypes_registry().get_datatype_by_extension(extension)
        if ret is None:
            log.warning("Datatype class not found for extension '%s'" % extension)
            return _get_datatypes_registry().get_datatype_by_extension('data')
        return ret
```

and this fixed my problem so the display code also works. I guess my counter question would be why enforce that it isn't ``None`` here? I'll pursue the Galaxy side fix but I think this is the right thing to do in bioblend.

jmchilton:
Closing in lieu of work around here: https://github.com/galaxyproject/planemo/pull/736. I needed to rework Planemo's download to be compatible with downstream changes anyway. Still think this should be merged though.

nsoranzo:
@jmchilton Sorry for dropping the ball on this, but you said "I'll pursue the Galaxy side fix", so I left this alone.

I've introduced the code you cited above in https://github.com/galaxyproject/galaxy/commit/0bfa1df72f8d0d140dea400ea6ad7f854714817e and you are right that `file_ext` (being mapped directly from `extension`) can be `None`, but can also be `'auto'` or `'_sniff_'`.

What do you modifying the BioBlend `download_dataset()` method by adding code similar to the one above:
```python
if not file_ext or file_ext == 'auto' or file_ext == '_sniff_':
    file_ext = 'data'
```

nsoranzo:
I've applied the above-mentioned change in commit f946b3ab5ef7362f0af34d0324b6087eb3d29d5f , I'm closing this PR. Thanks @jmchilton for reporting and the initial fix!

PR #249, opened 2017-09-07
Expose the requests timeout option for HTTP methods.

jmchilton:


nsoranzo:
Can you also add the `timeout` parameter to the `requests.get()` call in `bioblend/galaxy/datasets/__init__.py` , please?

jmchilton:
@nsoranzo Good catch (as always) - thanks. I've made change and updated the PR.

nsoranzo:
Thanks @jmchilton! FYI I'll drop Python 2.6 from `.travis.yml`, the new [wheel](https://pypi.python.org/pypi/wheel) 0.30.0 dropped support for it.

PR #251, opened 2017-09-14
Fix use of GalaxyClient with username/password on py3

mvdbeek:


bioblend/galaxyclient.py
```diff
@@ -201,7 +201,7 @@ def make_put_request(self, url, payload=None, params=None):
     @property
     def key(self):
         if not self._key and self.email is not None and self.password is not None:
-            unencoded_credentials = "%s:%s" % (self.email, self.password)
+            unencoded_credentials = b"%s:%s" % (self.email, self.password)
```
nsoranzo: This raises an exception for me:
`TypeError: unsupported operand type(s) for %: 'bytes' and 'tuple'`

what about modifying the next line instead to:
```python
authorization = base64.b64encode(unencoded_credentials.encode())
```

mvdbeek: I could have sworn I tested this under py3 as well :).

mvdbeek:
I've rebased the commit with the right fix.

nsoranzo:
Thanks @mvdbeek!

PR #256, opened 2018-02-01
add testing for Galaxy release 18.01

martenson:


nsoranzo:
We need to adapt the test configuration for the `galaxy.ini.sample` -> `galaxy.yml.sample` change.

PR #257, opened 2018-02-08
Use a template for the test galaxy.ini

nsoranzo:


nsoranzo:
Travis build failures on dev are unrelated, it seems now library uploads sometimes finish after temporary files are deleted.

PR #266, opened 2018-08-06
Allow retrying more than GET

erasche:
extracted retry logic from `_get` into a `retry` method.

- untested (even locally) just want to discuss implementation
- should retry be `_retry`?
- renames existing methods. Should the old names be retained as 'legacy' API? Should they include a deprecation notice somehow?

### DIFF
--- a/bioblend/galaxy/__init__.py
+++ b/bioblend/galaxy/__init__.py
@@ -71,20 +71,39 @@ class GalaxyInstance(GalaxyClient):
         self.folders = folders.FoldersClient(self)
 
     @property
-    def max_get_attempts(self):
-        return Client.max_get_retries()
+    def max_attempts(self):
+        return Client.max_retries()
 
-    @max_get_attempts.setter
-    def max_get_attempts(self, v):
-        Client.set_max_get_retries(v)
+    @max_attempts.setter
+    def max_attempts(self, v):
+        Client.set_max_retries(v)
 
     @property
-    def get_retry_delay(self):
-        return Client.get_retry_delay()
+    def retry_delay(self):
+        return Client.retry_delay()
 
-    @get_retry_delay.setter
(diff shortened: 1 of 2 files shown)


nuwang:
@erasche Nice! Perhaps the tenacity library could be used to simplify implementation?: https://github.com/jd/tenacity

erasche:
@nuwang I'm fond of https://github.com/litl/backoff for that as well.

Currently this just retains the existing implementation of retrying as it did not seem worth it to me to add a new dependency for this when the existing GET retrying logic works OK. But if that is preferred for bioblend's maintenance, we can certainly swap it for a library. 

I worry that it would only eliminate maybe 5-10 lines, I am under the impression we still need to retain the logic for setting the maximum values as client object attributes (not sure how that plays with decorators, never tried changing the definition of a decorator by using a variable set at runtime rather than hardcoded values,) and we still need to retain the behaviour of checking the exit code and attempting json conversion if requested

nuwang:
backoff looks great! Don't think I've come across it before.

Sure, makes sense if a lot of (future) variation in the retrying logic isn't expected. Calling the decorator manually with dynamic parameters is doable, but I guess it depends on what it buys, which may not be that much. I only suggested it because we've been using it in CloudBridge for some similar operations, and the library (we originally used retrying) has been fairly stable over time.

erasche:
It's a good suggestion! I personally prefer retry libraries for this since you can do nicer things like exponential backoff. Maybe @nsoranzo has an opinion when he's back from vacation

nsoranzo:
@erasche Can you take a look at the failing tests on Travis?

erasche:
@nsoranzo 

> untested (even locally) just want to discuss implementation

I didn't update any of the tests until I got your OK on the implementation. If you're happy with that I'll go ahead and fix the tests.

PR #267, opened 2018-08-30
Add missing api method/arg for data libraries

abretaud:
Hi,
Here's a little update to the data library api:
- add library dataset update method
- add `preserve_dirs` and `tag_using_filenames` arguments when loading data into a data library

### DIFF
--- a/bioblend/_tests/TestGalaxyLibraries.py
+++ b/bioblend/_tests/TestGalaxyLibraries.py
@@ -74,6 +74,13 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
         dataset_id = self._test_dataset(history['id'])
         self.gi.libraries.copy_from_dataset(self.library['id'], dataset_id, message='Copied from dataset')
 
+    def test_update_dataset(self):
+        library_id = self.library["id"]
+        dataset1_id = self._test_dataset(library_id)
+        updated_dataset = self.gi.libraries.update_library_dataset(library_id, dataset1_id, name='Modified name', misc_info='Modified the name succesfully')
+        self.assertEqual(updated_dataset["name"], 'Modified name')
+        self.assertEqual(updated_dataset["misc_info"], 'Modified the name succesfully')
+
     @test_util.skip_unless_galaxy('release_14.10')
     def test_library_permissions(self):
         current_user = self.gi.users.get_current_user()
--- a/bioblend/_tests/TestGalaxyLibraries.py
+++ b/bioblend/_tests/TestGalaxyLibraries.py
@@ -74,6 +74,7 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
         dataset_id = self._test_dataset(history['id'])
         self.gi.libraries.copy_from_dataset(self.library['id'], dataset_id, message='Copied from dataset')
 
+    @test_util.skip_unless_galaxy('release_17.09')
     def test_update_dataset(self):
         library_id = self.library["id"]
         dataset1 = self.gi.libraries.upload_file_contents(library_id, FOO_DATA)
--- a/bioblend/_tests/TestGalaxyLibraries.py
+++ b/bioblend/_tests/TestGalaxyLibraries.py
@@ -76,8 +76,8 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
 
     def test_update_dataset(self):
         library_id = self.library["id"]
-        dataset1_id = self._test_dataset(library_id)
-        updated_dataset = self.gi.libraries.update_library_dataset(dataset1_id, name='Modified name', misc_info='Modified the name succesfully')
+        dataset1 = self.gi.libraries.upload_file_contents(library_id, FOO_DATA)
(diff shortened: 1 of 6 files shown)


abretaud:
Thanks @nsoranzo!
Do you see an obvious reason for the failing test at https://travis-ci.org/galaxyproject/bioblend/jobs/422528629#L1622?

bioblend/galaxy/libraries/__init__.py
```diff
@@ -304,6 +335,10 @@ def _do_upload(self, library_id, **keywords):
             payload["roles"] = keywords["roles"]
         if keywords.get("link_data_only", None) and keywords['link_data_only'] != 'copy_files':
             payload["link_data_only"] = 'link_to_files'
+        if keywords.get('tag_using_filenames', False):
+            payload["tag_using_filenames"] = False
+        if keywords.get('preserve_dirs', True):
+            payload["preserve_dirs"] = True
```
nsoranzo: What about

```python
payload['tag_using_filenames'] = str(keywords.get('tag_using_filenames', True))
payload['preserve_dirs'] = str(keywords.get('preserve_dirs', False))
```

? The casting to string should solve the `MultipartEncoder` encode issue (in https://github.com/galaxyproject/bioblend/blob/master/bioblend/galaxyclient.py#L124 when there are files attached we don't json.dumps() the payload.

bioblend/galaxy/libraries/__init__.py
```diff
@@ -96,6 +96,37 @@ def delete_library_dataset(self, library_id, dataset_id, purged=False):
...
+
+        :type genome_build: str
+        :param genome_build: Replace library dataset genome build (dbkey)
+
+        :rtype: dict
+        :return: details of the updated dataset
+        """
+        url = self.gi._make_url(self, library_id, contents=True)
+        # Append the dataset_id to the base history contents URL
+        url = '/'.join([url, dataset_id])
```
nsoranzo: This will set `url` to `<base_url>/api/libraries/<library_id>/contents/<dataset_id>`, which executes this: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/library_contents.py#L373 , which is not the usual Galaxy `update` API, but has been there [since 2011](https://github.com/galaxyproject/galaxy/commit/38d26e946e6b0e2438ffc25e09923938860b5b35).

The `url` you want is `<base_url>/api/libraries/datasets/<dataset_id>` , i.e.:
```python
url = '/'.join([self.gi._make_url(self), 'datasets', dataset_id])
```

You can also remove `library_id` from the method parameters.

abretaud:
I get a 404 error with this url, do you think it comes from the PATCH http method on https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/library_datasets.py#L164?

nsoranzo:
@abretaud Yes, we need to add `GalaxyClient.make_patch_request()` and `Client._patch()` methods, I started having a look, but if you have time please go ahead, just let me know.

abretaud:
Ok, I jut a made a quick patch, we'll see if everything gets green now (it's greener locally :) )

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -74,6 +74,13 @@ def test_copy_from_dataset(self):
         dataset_id = self._test_dataset(history['id'])
         self.gi.libraries.copy_from_dataset(self.library['id'], dataset_id, message='Copied from dataset')
```
nsoranzo: Add `@test_util.skip_unless_galaxy('release_17.09')`

bioblend/_tests/TestGalaxyObjects.py
```diff
@@ -536,6 +536,16 @@ def test_dataset_delete(self):
         # by the API at the moment
         # self.assertTrue(self.ds.deleted)
```
nsoranzo: Add `@test_util.skip_unless_galaxy('release_17.09')`

nsoranzo:
Thanks @abretaud!

abretaud:
Thanks for the help!

bioblend/galaxy/libraries/__init__.py
```diff
@@ -304,6 +330,8 @@ def _do_upload(self, library_id, **keywords):
             payload["roles"] = keywords["roles"]
         if keywords.get("link_data_only", None) and keywords['link_data_only'] != 'copy_files':
             payload["link_data_only"] = 'link_to_files'
+        payload['tag_using_filenames'] = str(keywords.get('tag_using_filenames', True))
```
mvdbeek: I might be wrong here, but I think this shouldn't be a string ?

abretaud: Not sure... Have you seen this comment from @nsoranzo https://github.com/galaxyproject/bioblend/pull/267#discussion_r214043361 ?

nsoranzo: @mvdbeek @abretaud Thanks for bringing this up, I've corrected the JSON-encoding for Multipart POST requests in commit https://github.com/galaxyproject/bioblend/commit/c5059e550c54de7fd60a90a48bb627918e42eca7 .

PR #272, opened 2019-01-18
Pass BIOBLEND_TEST_JOB_TIMEOUT env var when testing

abretaud:
I had to increase the job timeout for travis tests of galaxy docker images (https://github.com/galaxyproject/galaxy-flavor-testing/pull/14 and https://github.com/bgruening/docker-galaxy-stable/commit/1cf32151926fa670f8b6e9f8e61e01b68f22c810#diff-c5e05a31e76f346b32fdb1f3a024e819), so I guess it's better to change it here


nsoranzo:
Thanks @abretaud !

Issue #239, opened 2017-07-07
get_libraries deleted flag not working properly


emartchenko:
Using bioblend version 0.9.0, galaxy release 17.05

get_libraries() returns all libraries, including those that have been deleted
get_libraries(deleted=False) also returns all libraries, including deleted ones
get_libraries(deleted=True) returns only deleted libraries

nsoranzo:
Fixed in #273. Now:
- `get_libraries()` and `get_libraries(deleted=False)` return only non-deleted libraries
- `get_libraries(deleted=True)` returns only deleted libraries (for admins only)
- `get_libraries(deleted=None)` returns all libraries (for admins only)

PR #273, opened 2019-01-24
fixed library retrieval behavior to reflect documentation, added tests

luke-c-sargent:
I encountered the issue mentioned in #239: the `deleted` flag for get_libraries() was not being properly handled.

the code [in libraries/get_libraries()](https://github.com/galaxyproject/bioblend/blob/c5059e550c54de7fd60a90a48bb627918e42eca7/bioblend/galaxy/libraries/__init__.py#L267) sets a `deleted` flag for `self._get()`, which [attempts to make the url with or without `deleted` as part of the URI path](https://github.com/galaxyproject/bioblend/blob/c5059e550c54de7fd60a90a48bb627918e42eca7/bioblend/galaxyclient.py#L68), not a URI parameter. however, in the case where deleted is False, this URI path component is not included, which gives all libraries including deleted libraries. That should only happen when deleted is set to None.

These changes send `deleted` as a parameter, which [is caught further in the `self._get()` method](https://github.com/galaxyproject/bioblend/blob/c5059e550c54de7fd60a90a48bb627918e42eca7/bioblend/galaxy/client.py#L107), and correctly returns the requested values.

- deleted = False -> does not include deleted
- deleted = True -> only includes deleted
- deleted = None -> includes both

the added tests verify this functionality.

### DIFF
--- a/bioblend/_tests/TestGalaxyLibraries.py
+++ b/bioblend/_tests/TestGalaxyLibraries.py
@@ -12,7 +12,10 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
     def setUp(self):
         super(TestGalaxyLibraries, self).setUp()
         self.name = 'automated test library'
+        self.deleted_name = 'deleted test library'
         self.library = self.gi.libraries.create_library(self.name, description='automated test', synopsis='automated test synopsis')
+        self.deleted_library = self.gi.libraries.create_library(self.deleted_name, description='a deleted library', synopsis='automated test synopsis')
+        self.gi.libraries.delete_library(self.deleted_library['id'])
 
     def tearDown(self):
         self.gi.libraries.delete_library(self.library['id'])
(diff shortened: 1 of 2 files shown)


bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -12,7 +12,10 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
     def setUp(self):
         super(TestGalaxyLibraries, self).setUp()
         self.name = 'automated test library'
+        self.deleted_name = 'deleted test library'
         self.library = self.gi.libraries.create_library(self.name, description='automated test', synopsis='automated test synopsis')
+        self.deleted_library = self.gi.libraries.create_library(self.deleted_name, description='a deleted library', synopsis='automated test synopsis')
+        self.gi.libraries.delete_library(self.deleted_library['id'])
```
nsoranzo: I'd move these 3 new lines to the `test_get_libraries()` method, since the deleted library is used only for that test.

nsoranzo:
Thanks @luke-c-sargent for the fix and tests!

I've made some changes to the new tests, hope you are OK with that.
Tests on Python2 are failing because of https://github.com/jaraco/configparser/pull/24/files#r250956907 , let's see if this is resolved rapidly upstream.

luke-c-sargent:
thanks @nsoranzo for checking it out / improving it so quickly!

PR #283, opened 2019-05-08
Fix type parameter in HistoryClient.show_history

mvdbeek:


bioblend/_tests/TestGalaxyHistories.py
```diff
@@ -68,6 +68,22 @@ def test_show_history(self):
...
+        contents = self.gi.histories.show_history(self.history['id'], contents=True)
+        # Empty history has no datsets, content length should be 0
+        self.assertEqual(len(contents), 0)
+        self._test_dataset(history_id)
+        # history has 1 datset, content length should be 1
+        self.assertEqual(len(contents), 1)
+        contents = self.gi.histories.show_history(self.history['id'], contents=True, types=['dataset'])
+        # filtering for dataset, content length should still be 1
+        self.assertEqual(len(contents), 1)
+        contents = self.gi.histories.show_history(self.history['id'], contents=True, types=['dataset_collection'])
```
erasche: maybe add a test for `types=['dataset', 'dataset_collection']`?

nsoranzo:
@mvdbeek I've added a commit updating some docs, I hope that's OK.

bioblend/galaxy/histories/__init__.py
```diff
@@ -76,23 +76,28 @@ def show_history(self, history_id, contents=False, deleted=None, visible=None, d
         :param history_id: Encoded history ID to filter on
 
         :type contents: bool
-        :param contents: When ``True``, the complete list of datasets in the
-          given history.
+        :param contents: When ``True``, return also the complete list of
```
mvdbeek: It's not `also`, it's a different output altogether.
```suggestion
        :param contents: When ``True``, return the complete list of
```

mvdbeek:
Sure, no problem.

bioblend/galaxy/histories/__init__.py
nsoranzo: Uh, that's right! I've made it even more explicit in a rebase of the last commit.

PR #291, opened 2019-07-05
Documentation on 'raw' API usage through bioblend

selten:
Add documentation on how to use bioblend to ship the request to the Galaxy API. This could be used to avoid having both requests and bioblend used in a single script that needs administrative privileges to access Galaxy or does API transactions that bioblend does not yet support.


nsoranzo:
@selten Thanks for the nice addition! I've add a commit trying to clarify a bit, hope it's fine with you.

selten:
@nsoranzo Great! I'm not that good in writing docs. Your changes improve them greatly!

PR #292, opened 2019-07-12
make range in wait_for_dataset use int

pvanheus:
The signature of `wait_for_dataset` in the `libraries` module suggests the `maxwait` and `interval` parameters are of type float. These are then used in a `range()` function which expects `int` types, thus triggering an error. This patch forces them to integers.

### DIFF
--- a/bioblend/galaxy/libraries/__init__.py
+++ b/bioblend/galaxy/libraries/__init__.py
@@ -168,7 +168,7 @@ class LibraryClient(Client):
         assert maxwait > 0
         assert interval > 0
 
-        for time_left in range(maxwait, 0, -interval):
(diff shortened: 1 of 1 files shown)


nsoranzo:
Sorry for this issue I introduced in https://github.com/galaxyproject/bioblend/pull/257 . The similar method `DatasetClient._block_until_dataset_terminal()` does support float `maxwait` and `interval` (funnily thanks to the changes in the same PR), so I'm going to fix this in the same way, if that's OK: https://github.com/nsoranzo/bioblend/commit/4ef8e413b8da214da0dd6a2476ca98cef31e2d12

nsoranzo:
Thanks @pvanheus for reporting the issue!

PR #296, opened 2019-08-28
Add support for tagging library datasets

abretaud:
This is the corresponding changes to the API change in https://github.com/galaxyproject/galaxy/pull/8530

### DIFF
--- a/bioblend/_tests/TestGalaxyLibraries.py
+++ b/bioblend/_tests/TestGalaxyLibraries.py
@@ -126,33 +126,6 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
         self.assertEqual(dataset_show['tags'], 'name:foobar, name:barfoo')
 
     @test_util.skip_unless_galaxy('release_19.09')
-    def test_upload_from_galaxy_filesystem_with_tags(self):
-        bnames = ['f%d.txt' % i for i in range(2)]
-        tempdir = tempfile.mkdtemp(prefix='bioblend_test_')
-        try:
-            fnames = [os.path.join(tempdir, _) for _ in bnames]
-            for fn in fnames:
-                with open(fn, 'w') as f:
-                    f.write(FOO_DATA)
-            filesystem_paths = '\n'.join(fnames)
-            ret = self.gi.libraries.upload_from_galaxy_filesystem(self.library['id'], filesystem_paths, tags=["foobar", "barfoo"])
-            for dataset_dict in ret:
-                dataset = self.gi.libraries.wait_for_dataset(self.library['id'], dataset_dict['id'])
-                self.assertEqual(dataset['state'], 'ok')
-
-                dataset_show = self.gi.libraries.show_dataset(self.library['id'], dataset_dict['id'])
-                self.assertEqual(dataset_show['tags'], 'name:foobar, name:barfoo')
-            ret = self.gi.libraries.upload_from_galaxy_filesystem(self.library['id'], filesystem_paths, link_data_only='link_to_files', tags=["foobar", "barfoo"])
-            for dataset_dict in ret:
-                dataset = self.gi.libraries.wait_for_dataset(self.library['id'], dataset_dict['id'])
-                self.assertEqual(dataset['state'], 'ok')
-
-                dataset_show = self.gi.libraries.show_dataset(self.library['id'], dataset_dict['id'])
-                self.assertEqual(dataset_show['tags'], 'name:foobar, name:barfoo')
-        finally:
-            shutil.rmtree(tempdir)
-
-    @test_util.skip_unless_galaxy('release_19.09')
     def test_update_dataset_tags(self):
         datasets = self.gi.libraries.upload_file_contents(self.library['id'], FOO_DATA)
         dataset_show = self.gi.libraries.show_dataset(self.library['id'], datasets[0]['id'])
--- a/bioblend/_tests/TestGalaxyLibraries.py
+++ b/bioblend/_tests/TestGalaxyLibraries.py
@@ -45,7 +45,7 @@ class TestGalaxyLibraries(GalaxyTestBase.GalaxyTestBase):
         self.assertEqual(self.library['name'], library_data['name'])
 
     def test_upload_file_from_url(self):
-        pass
+        self.gi.libraries.upload_file_from_url(self.library['id'], 'https://zenodo.org/record/582600/files/wildtype.fna?download=1')
 
     def test_upload_file_contents(self):
         self.gi.libraries.upload_file_contents(self.library['id'], FOO_DATA)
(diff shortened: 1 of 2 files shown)


bioblend/galaxy/libraries/__init__.py
```diff
@@ -117,6 +117,9 @@ def update_library_dataset(self, dataset_id, **kwds):
         :type genome_build: str
         :param genome_build: Replace library dataset genome build (dbkey)
 
+        :type tags: list
+        :param tags: A list of tags to add to the dataset
```
nsoranzo: Can this also be used to remove tags?

bioblend/galaxy/libraries/__init__.py
```diff
@@ -344,6 +347,8 @@ def _do_upload(self, library_id, **keywords):
         if keywords.get("link_data_only", None) and keywords['link_data_only'] != 'copy_files':
             payload["link_data_only"] = 'link_to_files'
         payload['tag_using_filenames'] = keywords.get('tag_using_filenames', True)
+        if keywords.get('tags', None) is not None:
```
nsoranzo:
```suggestion
        if keywords.get('tags'):
```

bioblend/galaxy/libraries/__init__.py
abretaud: Actually it replaces existing tags with the ones given, I will change the docstring

abretaud: (and yes, I'll add tests!)

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
         self.assertEqual(set(_[1] for _ in ret_get['access_dataset_roles']), set(user_id_list_new))
         self.assertEqual(set(_[1] for _ in ret_get['modify_item_roles']), set(user_id_list_new))
         self.assertEqual(set(_[1] for _ in ret_get['manage_dataset_roles']), set(user_id_list_new))
+
+    @test_util.skip_unless_galaxy('release_19.09')
+    def test_upload_file_contents_with_tags(self):
+        with tempfile.NamedTemporaryFile(mode='w', prefix='bioblend_test_') as f:
```
nsoranzo:
```suggestion
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+            f.write(FOO_DATA)
```
nsoranzo:
```suggestion
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+            f.flush()
```
nsoranzo:
```suggestion
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+            datasets = self.gi.libraries.upload_file_contents(self.library['id'], FOO_DATA, tags=["foobar", "barfoo"])
```
nsoranzo:
```suggestion
        datasets = self.gi.libraries.upload_file_contents(self.library['id'], FOO_DATA, tags=["foobar", "barfoo"])
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+            dataset_show = self.gi.libraries.show_dataset(self.library['id'], datasets[0]['id'])
```
nsoranzo:
```suggestion
        dataset_show = self.gi.libraries.show_dataset(self.library['id'], datasets[0]['id'])
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+            self.assertEqual(dataset_show['tags'], 'name:foobar, name:barfoo')
```
nsoranzo:
```suggestion
        self.assertEqual(dataset_show['tags'], 'name:foobar, name:barfoo')
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+
+    @test_util.skip_unless_galaxy('release_19.09')
```
nsoranzo: Given that all the upload methods just call `_do_upload()`, I think just testing tags with the `test_upload_file_contents_with_tags` test above is enough?
So `test_upload_file_local_path_with_tags` and `test_upload_from_galaxy_filesystem_with_tags` are not needed, IMO.

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+                dataset = self.gi.libraries.wait_for_dataset(self.library['id'], dataset_dict['id'])
+                self.assertEqual(dataset['state'], 'ok')
+
+                dataset_show = self.gi.libraries.show_dataset(self.library['id'], dataset_dict['id'])
+                self.assertEqual(dataset_show['tags'], 'name:foobar, name:barfoo')
+        finally:
+            shutil.rmtree(tempdir)
+
+    @test_util.skip_unless_galaxy('release_19.09')
+    def test_upload_file_local_path_update_tags(self):
```
nsoranzo:
```suggestion
    def test_update_dataset_tags(self):
```

bioblend/_tests/TestGalaxyLibraries.py
```diff
@@ -118,3 +118,62 @@ def test_dataset_permissions(self):
...
+        with tempfile.NamedTemporaryFile(mode='w', prefix='bioblend_test_') as f:
+            f.write(FOO_DATA)
+            f.flush()
+            datasets = self.gi.libraries.upload_file_from_local_path(self.library['id'], f.name)
```
nsoranzo: I'd use the faster `upload_file_contents` method as in `test_update_dataset()` above.

bioblend/_tests/TestGalaxyLibraries.py
abretaud: In fact tags are not added the same way between upload_from_galaxy_filesystem and the other methods. I wrote the 3 tests to understand why sometimes the tags were not added.
Now that it works for all, I removed test_upload_file_local_path_with_tags as it's clearly redundant, do you prefer to remove test_upload_from_galaxy_filesystem_with_tags also?

abretaud:
I think the tests should pass now as https://github.com/galaxyproject/galaxy/pull/8530 was merged, maybe a rerun (or it can wait the 19.09 release)?

bioblend/_tests/TestGalaxyLibraries.py
nsoranzo: I think that tagging is now ensured for all code paths by https://github.com/galaxyproject/galaxy/pull/8530/files#diff-75e29f5ca09c0d8c3d9ef0286e1483dfR284-R286 , so I'd remove also `test_upload_from_galaxy_filesystem_with_tags` .

nsoranzo:
Thanks @abretaud !

PR #300, opened 2019-11-12
Get job metrics

AndreasSko:
Added function to get the job metrics for a specified job. `show_job` offers this functionality in part, however the Galaxy API in that case only returns metrics if the user is an admin. This call allows to fetch metrics even as a normal user as long as the Galaxy instance has `expose_potentially_sensitive_job_metrics` set to true.

### DIFF
--- a/bioblend/galaxy/jobs/__init__.py
+++ b/bioblend/galaxy/jobs/__init__.py
@@ -119,6 +119,13 @@ class JobsClient(Client):
         """
         Return job metrics for specified job.
 
+        .. note::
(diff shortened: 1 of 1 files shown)


nsoranzo:
Thanks a lot for the pull request! Can you please add the info about `expose_potentially_sensitive_job_metrics` to the method docstring, similarly to what we have e.g. in `bioblend/galaxy/users/__init__.py`?

AndreasSko:
Added a note :)

nsoranzo:
Thanks @AndreasSko !

PR #302, opened 2019-11-15
Add two new functions for checking invocations

simonbray:
Functions corresponding to `/api/workflows/{workflow_id}/invocations/{invocation_id}/jobs_summary` and `/api/workflows/{workflow_id}/invocations/{invocation_id}/report`.

### DIFF
--- a/bioblend/galaxy/invocations/__init__.py
+++ b/bioblend/galaxy/invocations/__init__.py
@@ -10,10 +10,28 @@ class InvocationClient(Client):
         self.module = 'invocations'
         super(InvocationClient, self).__init__(galaxy_instance)
 
+    def get_invocations(self):
+        """
+        Get a list containing all workflow invocations.
+
+        :rtype: list
+        :return: A list of workflow invocations.
+          For example::
+
+            [{u'history_id': u'2f94e8ae9edff68a',
+              u'id': u'df7a1f0c02a5b08e',
+              u'model_class': u'WorkflowInvocation',
+              u'state': u'new',
+              u'update_time': u'2015-10-31T22:00:22',
+              u'uuid': u'c8aa2b1c-801a-11e5-a9e5-8ca98228593c',
+              u'workflow_id': u'03501d7626bd192f'}]
+        """
+        return self._get()
+
     def show_invocation(self, invocation_id):
         """
-        Get a workflow invocation object representing the scheduling of a
-        workflow. This object may be sparse at first (missing inputs and
+        Get a workflow invocation dictionary representing the scheduling of a
+        workflow. This dictionary may be sparse at first (missing inputs and
         invocation steps) and will become more populated as the workflow is
         actually scheduled.
 
@@ -56,31 +74,7 @@ class InvocationClient(Client):
              u'workflow_id': u'03501d7626bd192f'}
         """
         url = self._invocation_url(invocation_id)
-        # return self._get(url=url)
-        # url = self.gi._make_url(self)
-        # url = _join(url, invocation_id)
         return self._get(url=url)
-
-    def get_invocations(self):
-        """
-        Get a list containing all the workflow invocations corresponding to the
-        specified workflow.
-
-        :rtype: list
-        :return: A list of workflow invocations.
-          For example::
-
-            [{u'history_id': u'2f94e8ae9edff68a',
-              u'id': u'df7a1f0c02a5b08e',
-              u'model_class': u'WorkflowInvocation',
-              u'state': u'new',
-              u'update_time': u'2015-10-31T22:00:22',
-              u'uuid': u'c8aa2b1c-801a-11e5-a9e5-8ca98228593c',
-              u'workflow_id': u'03501d7626bd192f'}]
-        """
-        # url = _join(self.gi._make_url(self), "invocations")
-        # return self._get(url=url)
-        return self._get()
 
     def cancel_invocation(self, invocation_id):
         """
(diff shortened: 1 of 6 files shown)


simonbray:
ping @tdudgeon

bgruening:
@simonbray shouldn't `https://usegalaxy.eu/api/invocations/<invocation_id>/jobs_summary` also work? So we do not need to workflow id?

simonbray:
Yes, both work. However, as far as I can see all the invocation-related functions use the former version (taking both workflow and invocation ids) so I just stuck with that.

bgruening:
@jmchilton what is the way to go here. My gut feeling is that specifying the workflow ID if it is not needed is unnesesary complex.

jmchilton:
The workflow ID is not needed in 19.09 after https://github.com/galaxyproject/galaxy/pull/8664/files.

simonbray:
Okay, will remove it.

simonbray:
Actually there are multiple options:
1) just remove the workflow ID parameter from the functions I added here, which introduces inconsistency within the usage of invocation-related functions.
2) rewrite all the invocation functions (including `get_invocations`, `show_invocation` etc) to remove the workflow ID parameter. It would be more consistent and simpler but I guess this has the disadvantage that they will no longer be accessible for pre-19.09 versions.
3) rewrite all functions but allow the workflow ID to be optionally specified for users of older versions.

Not sure which is best, it depends how much weight is given to catering to users of older versions. Please comment @bgruening or @jmchilton.

nsoranzo:
@simonbray I'd suggest to:
- add a new module `bioblend/galaxy/invocations/__init__.py`
- move the methods that work only since 19.09 to the new module and remove the workflow ID parameter from them
- copy the methods that do not require the workflow ID any more from `workflows/__init__.py` to the new module, again removing the unnecessary parameter.

bgruening:
All green, thanks a lot @simonbray. Is this still WIP?

simonbray:
Not as far as I'm concerned

bgruening:
Great. @nsoranzo can you have a look at it. We would need this for a few downstream projects.

Thanks!

bioblend/galaxy/invocations/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+        url = self._invocation_url(invocation_id)
+        # return self._get(url=url)
+        # url = self.gi._make_url(self)
+        # url = _join(url, invocation_id)
+        return self._get(url=url)
+
+    def get_invocations(self):
+        """
+        Get a list containing all the workflow invocations corresponding to the
+        specified workflow.
```
nsoranzo: There is no "specified workflow", so I guess this docstring needs to be updated.

bioblend/galaxy/invocations/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+             u'uuid': u'c8aa2b1c-801a-11e5-a9e5-8ca98228593c',
+             u'workflow_id': u'03501d7626bd192f'}
+        """
+        url = self._invocation_url(invocation_id)
+        # return self._get(url=url)
+        # url = self.gi._make_url(self)
+        # url = _join(url, invocation_id)
+        return self._get(url=url)
+
+    def get_invocations(self):
```
nsoranzo: I'd put this method before `show_invocation()` for consistency with the other modules.

bioblend/galaxy/invocations/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+        the type of workflow step (the only currently valid action is True/False
+        for pause steps).
+
+        :type invocation_id: str
+        :param invocation_id: Encoded workflow invocation ID
+
+        :type step_id: str
+        :param step_id: Encoded workflow invocation step ID
+
+        :type action: object
```
nsoranzo: Is `action` a dictionary or?
It would help to have a test :)

bioblend/galaxy/invocations/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+        :rtype: dict
+        :return: Representation of the workflow invocation step
+        """
+        url = self._invocation_step_url(invocation_id, step_id)
+        payload = {"action": action}
+        return self._put(payload=payload, url=url)
+
+    def get_invocation_summary(self, invocation_id):
+        """
+        Get a summary of an invokation, stating the number of jobs which
```
nsoranzo:
```suggestion
        Get a summary of an invocation, stating the number of jobs which
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+             'model': 'WorkflowInvocation',
+             'id': 'a799d38679e985db',
+             'populated_state': 'ok'}
+        """
+        url = _join(self._invocation_url(invocation_id), "jobs_summary")
+        return self._get(url=url)
+
+    def get_invocation_report(self, invocation_id):
+        """
+        Get a markdown report for an invokation.
```
nsoranzo:
```suggestion
        Get a Markdown report for an invocation.
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -0,0 +1,203 @@
...
+               u'workflow_step_label': None,
+               u'workflow_step_uuid': u'e62440b8-e911-408b-b124-e05435d3125e'}],
+             u'update_time': u'2015-10-31T22:00:26',
+             u'uuid': u'c8aa2b1c-801a-11e5-a9e5-8ca98228593c',
+             u'workflow_id': u'03501d7626bd192f'}
+        """
+        url = self._invocation_url(invocation_id)
+        # return self._get(url=url)
+        # url = self.gi._make_url(self)
+        # url = _join(url, invocation_id)
```
nsoranzo: Can you remove commented-out lines here too?

nsoranzo:
Thanks a lot for the cool additions @simonbray !

PR #312, opened 2020-01-20
Drop support for Python 2.7. Add support for Galaxy release_20.01

nsoranzo:


jmchilton:
-.5 on dropping support for Python given that Planemo and Galaxy still depend on this. I think Python 2 is a lot more of a reasonable target than some of the older Galaxies this project tries to support.

nsoranzo:
For context, my aim is to align our ecosystem to https://python3statement.org/

To be fair, the oldest Galaxy release BioBlend supports is 15.03 , and Python 3.5.0 (the proposed new minimum Python version) was also released in 2015.

I'd like to also drop Python 2 support for Planemo (so that won't be an issue), and Galaxy can pin BioBlend (like it needs to do with at least other 5 now Python3-only dependencies) as long as we want to keep Python2 support for it.

jmchilton:
To be fair I'd also be -.5 on dropping Python 2 support for Planemo. It seems needless, we're breaking backward compatibility without a killer feature, dependency or use case planned to justify it. I've always found https://python3statement.org/ gross. If you don't want to support Python 2 because it is hard for you or not worth the effort, that is fine and totally justified there are so many cool things Python 3 only projects can do but to wrap hurting accessibility up as some noble effort that is helping the ecosystem seems gross to me.

drosofff:
Agree with @jmchilton 👍 To be fair I am not working (yet) on planemo and my voice does not count. But I know the issue because ansible-playbooks use all these nice codes which are impacted by the python migration. The smoother transition the better

mvdbeek:
I feel like I'm working around python2 shortcomings and making code more complicated / less readable than necessary every day now. It needs to go. killer features aren't needed IMO, we didn't have any in the python 2.6 to 2.7 transition either. There's just such a huge amount of small niceties (not just for developers, deployers will benefit from many small improvements, such as non-broken cryptography as well.) we shouldn't be living without.

jmchilton:
Those seem to be reasons you dislike Python 2 for Galaxy development - nothing to do with delivering features or Bioblend. I stand by everything I said.

mvdbeek:
That is true, thanks for the merge @jmchilton!

martenson:
Fwiw I think supporting py2 does not cost bioblend much (if anything) so
changing it just for sake of being "proper" feels wrong to me.

PR #318, opened 2020-01-29
Allow specifying the User-Agent when creating a GalaxyClient

blankenberg:


nsoranzo:
@blankenberg Thanks for the PR! Can I ask what's the use case? A particular issue this is meant to solve?

hexylena:
Came to file the associated bug report, only to discover this PR!

I can provide a use case: identifying who i am in server access. 

When I decide to go hammer the galaxy or toolshed API, it's generally courteous to self-identify and provide e.g. a github url so sysadmins know where to submit bug reports if they're experiencing undue load from my client. Nice bots share who they are, so, it's easier to communicate with their developers.

In some of my other [non-bioblend scripts](https://github.com/hexylena/toolshed-version-database/blob/main/fetch2.py#L37) I send a project name/version/github URL in the user-agent, so I'm fully in support of this and would love to see it rebased and available!

nsoranzo:
Thanks @hexylena , I've rebased the PR, can you have a look?

hexylena:
Apologies, missed the ping. Yes, this looks great.

nsoranzo:
> Apologies, missed the ping. Yes, this looks great.

Thanks for checking!

PR #319, opened 2020-01-31
Add an update_user() method to UserClient

abretaud:
Hi!
I needed a way to update user emails using the api, so here it is in bioblend

### DIFF
--- a/bioblend/_tests/TestGalaxyUsers.py
+++ b/bioblend/_tests/TestGalaxyUsers.py
@@ -76,7 +76,7 @@ class TestGalaxyUsers(GalaxyTestBase.GalaxyTestBase):
         updated_user_email = 'updateduser@example.com'
         updated_username = 'updateduser'
         user_id = user['id']
-        user = self.gi.users.update_user(user_id, username=updated_username, email=updated_user_email)
+        self.gi.users.update_user(user_id, username=updated_username, email=updated_user_email)
         user = self.gi.users.show_user(user_id)
         self.assertEqual(user['username'], updated_username)
         self.assertEqual(user['email'], updated_user_email)
--- a/bioblend/_tests/TestGalaxyUsers.py
+++ b/bioblend/_tests/TestGalaxyUsers.py
@@ -75,7 +75,7 @@ class TestGalaxyUsers(GalaxyTestBase.GalaxyTestBase):
 
         updated_user_email = 'updateduser@example.com'
         updated_username = 'updateduser'
-        current_user = user['id']
+        user_id = user['id']
         user = self.gi.users.update_user(current_user, username=updated_username, email=updated_user_email)
         user = self.gi.users.show_user(current_user)
         self.assertEqual(user['username'], updated_username)
--- a/bioblend/_tests/TestGalaxyUsers.py
+++ b/bioblend/_tests/TestGalaxyUsers.py
@@ -61,3 +61,25 @@ class TestGalaxyUsers(GalaxyTestBase.GalaxyTestBase):
         self.assertIsNotNone(user['email'])
         self.assertIsNotNone(user['nice_total_disk_usage'])
         self.assertIsNotNone(user['total_disk_usage'])
+
+    def test_update_user(self):
+        # WARNING: only admins can create users!
+        # WARNING: Users cannot be purged through the Galaxy API, so execute
+        # this test only on a disposable Galaxy instance!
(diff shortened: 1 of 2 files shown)


bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -61,3 +61,27 @@ def test_get_current_user(self):
...
+        current_user = user['id']
+        user = self.gi.users.update_user(current_user, username=updated_username, email=updated_user_email)
+        user = self.gi.users.show_user(current_user)
+        self.assertEqual(user['username'], updated_username)
+        self.assertEqual(user['email'], updated_user_email)
+
+        if self.gi.config.get_config()['allow_user_deletion']:
+            deleted_user = self.gi.users.delete_user(user['id'])
+            self.assertEqual(deleted_user['email'], new_user_email)
+            self.assertTrue(deleted_user['deleted'])
```
nsoranzo: The 2 asserts are probably not needed.

bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -61,3 +61,27 @@ def test_get_current_user(self):
...
+        new_user_email = 'newuser2@example.com'
+        user = self.gi.users.create_local_user('newuser2', new_user_email, 'secret')
+        self.assertEqual(user['username'], 'newuser2')
+        self.assertEqual(user['email'], new_user_email)
+
+        updated_user_email = 'updateduser@example.com'
+        updated_username = 'updateduser'
+        current_user = user['id']
+        user = self.gi.users.update_user(current_user, username=updated_username, email=updated_user_email)
+        user = self.gi.users.show_user(current_user)
```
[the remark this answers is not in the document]
nsoranzo: Also can you just check the `user` dict returned by `self.gi.users.update_user()` ?

bioblend/galaxy/users/__init__.py
```diff
@@ -186,3 +186,23 @@ def get_user_apikey(self, user_id):
...
+        :type username: str
+        :param username: Replace user name with the given string
+
+        :type email: str
+        :param email: Replace user email with the given string
+
+        :rtype: dict
+        :return: details of the updated user
+        """
+        url = self._make_url(user_id) + '/information/inputs'
```
nsoranzo: I see there is also a `PUT /api/users/{id}` endpoint (`update()` method in `lib/galaxy/webapps/galaxy/api/users.py`). I suppose this allows to change more attributes?

bioblend/_tests/TestGalaxyUsers.py
nsoranzo: Any way the first assert is wrong, the 2nd parameter should be `updated_user_email`

bioblend/_tests/TestGalaxyUsers.py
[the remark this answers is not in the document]
abretaud: It's not returning the user info, just `{'message': 'User information has been saved.'}`

bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -61,3 +61,27 @@ def test_get_current_user(self):
...
+            self.skipTest('This Galaxy instance is not configured to use local users')
+        new_user_email = 'newuser2@example.com'
+        user = self.gi.users.create_local_user('newuser2', new_user_email, 'secret')
+        self.assertEqual(user['username'], 'newuser2')
+        self.assertEqual(user['email'], new_user_email)
+
+        updated_user_email = 'updateduser@example.com'
+        updated_username = 'updateduser'
+        user_id = user['id']
+        user = self.gi.users.update_user(user_id, username=updated_username, email=updated_user_email)
```
nsoranzo:
```suggestion
        self.gi.users.update_user(user_id, username=updated_username, email=updated_user_email)
```

bioblend/galaxy/users/__init__.py
abretaud: I just tried with `PUT /api/users/{id}`, but no change is done in the db... Maybe the payload needs some special format?

nsoranzo: It seems only updating the username is supported in https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/users.py#L586-L590

I think we should move the logic of user update from `UserAPIController.set_information()` to `UserDeserializer`, but for now this looks fine.

PR #320, opened 2020-01-31
Fixup workflow

### LINKED ISSUE #311: get_invocations(), forward history_id param
gi.workflows.get_invocations() should forward the optional history_id parameter to allow a reverse lookup of invocation from history.

innovate-invent:
Some of the documentation is incomplete as I am not sure what these parameters fully do.

Fix #311.

### DIFF
--- a/bioblend/galaxy/invocations/__init__.py
+++ b/bioblend/galaxy/invocations/__init__.py
@@ -10,9 +10,23 @@ class InvocationClient(Client):
         self.module = 'invocations'
         super().__init__(galaxy_instance)
 
-    def get_invocations(self):
+    def get_invocations(self, history_id=None, user_id=None, include_terminal=True, limit=None, view='collection', step_details=False):
         """
-        Get a list containing all workflow invocations.
+        Get a list containing all the workflow invocations corresponding to the
+        specified workflow.
+
+        :type history_id: str
+        :param history_id: Encoded history ID
+        :type user_id: str
+        :param user_id: TODO
+        :type include_terminal: bool
+        :param include_terminal: TODO
+        :type limit: int
+        :param limit: TODO
+        :type view: str
+        :param view: TODO
+        :type step_details: bool
+        :param step_details: include step details
 
         :rtype: list
         :return: A list of workflow invocations.
@@ -26,7 +40,14 @@ class InvocationClient(Client):
               'uuid': 'c8aa2b1c-801a-11e5-a9e5-8ca98228593c',
               'workflow_id': '03501d7626bd192f'}]
         """
-        return self._get()
+        params = {'include_terminal': include_terminal, 'view': view, 'step_details': step_details}
+        if history_id:
+            params['history_id'] = history_id
+        if user_id:
+            params['user_id'] = user_id
+        if limit:
+            params['limit'] = limit
+        return self._get(params=params)
 
     def show_invocation(self, invocation_id):
         """
--- a/bioblend/galaxy/objects/wrappers.py
+++ b/bioblend/galaxy/objects/wrappers.py
@@ -184,7 +184,7 @@ class Workflow(Wrapper):
     results from an input dataset.
     """
     BASE_ATTRS = Wrapper.BASE_ATTRS + (
-        'deleted', 'inputs', 'published', 'steps', 'tags', 'owner', 'number_of_steps', 'show_in_tool_panel', 'latest_workflow_uuid'
+        'deleted', 'inputs', 'published', 'steps', 'tags', 'owner', 'latest_workflow_uuid'
     )
     POLLING_INTERVAL = 10  # for output state monitoring
 
@@ -1453,7 +1453,7 @@ class WorkflowPreview(Preview):
     Instances of this class wrap dictionaries obtained by getting
     ``/api/workflows`` from Galaxy.
     """
-    BASE_ATTRS = Preview.BASE_ATTRS + ('published', 'tags')
+    BASE_ATTRS = Preview.BASE_ATTRS + ('published', 'tags', 'number_of_steps', 'show_in_tool_panel')
 
     def __init__(self, pw_dict, gi=None):
         super().__init__(pw_dict, gi=gi)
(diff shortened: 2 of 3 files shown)


bioblend/galaxy/objects/wrappers.py
```diff
@@ -193,7 +193,7 @@ class Workflow(Wrapper):
     results from an input dataset.
     """
     BASE_ATTRS = Wrapper.BASE_ATTRS + (
-        'deleted', 'inputs', 'published', 'steps', 'tags'
+        'deleted', 'inputs', 'published', 'steps', 'tags', 'owner', 'number_of_steps', 'show_in_tool_panel', 'latest_workflow_uuid'
```
nsoranzo: I think `number_of_steps` and `show_in_tool_panel` are not presently available in `/api/workflows/<workflow_id>` , but only in `/api/workflows/`, which means that they can only be added to `BASE_ATTRS` of the `WorkflowPreview` class.

bioblend/galaxy/workflows/__init__.py
```diff
@@ -632,13 +632,25 @@ def show_invocation(self, workflow_id, invocation_id):
...
         """
         Get a list containing all the workflow invocations corresponding to the
         specified workflow.
 
         :type workflow_id: str
         :param workflow_id: Encoded workflow ID
+        :type history_id: str
+        :param history_id: Encoded history ID
+        :type user_id: str
+        :param user_id: TODO
```
nsoranzo: Are you planning to fill up the TODOs, so still WIP?

innovate-invent: I'll need a dev or someone more familiar with the api to do this.

bioblend/galaxy/objects/wrappers.py
innovate-invent: Fixed

nsoranzo:
Sorry, I forgot that in https://github.com/galaxyproject/bioblend/pull/302 we introduced an `InvocationClient.get_invocations()` method in `bioblend/galaxy/invocations/__init__.py` which seems to be a better endpoint to extend. Can you move the added parameters there?

```python
    def get_invocations(self, workflow_id=None, history_id=None, user_id=None, include_terminal=True, limit=None, view='collection', step_details=False):
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -10,9 +10,23 @@ def __init__(self, galaxy_instance):
         self.module = 'invocations'
         super().__init__(galaxy_instance)
 
-    def get_invocations(self):
+    def get_invocations(self, history_id=None, user_id=None, include_terminal=True, limit=None, view='collection', step_details=False):
```
nsoranzo:
```suggestion
    def get_invocations(self, workflow_id=None, history_id=None, user_id=None, include_terminal=True, limit=None, view='collection', step_details=False):
```

Also need to document and use `workflow_id` below.

bioblend/galaxy/invocations/__init__.py
```diff
@@ -10,9 +10,23 @@ def __init__(self, galaxy_instance):
...
         """
-        Get a list containing all workflow invocations.
+        Get a list containing all the workflow invocations corresponding to the
+        specified workflow.
```
nsoranzo: Maybe "Select a list of workflow invocations"?

bioblend/galaxy/workflows/__init__.py
```diff
@@ -632,13 +632,25 @@ def show_invocation(self, workflow_id, invocation_id):
         url = self._invocation_url(workflow_id, invocation_id)
         return self._get(url=url)
 
-    def get_invocations(self, workflow_id):
+    def get_invocations(self, workflow_id, history_id=None, user_id=None, include_terminal=True, limit=None, view='collection', step_details=False):
```
nsoranzo: What about leaving this method as it is and add a note here suggesting to use `InvocationClient.get_invocations()` for more advanced filtering?

innovate-invent: is there a reason we shouldn't provide it?

bioblend/galaxy/invocations/__init__.py
innovate-invent: workflow_id wasn't specified in the original function, is it a url parameter?

nsoranzo: No, it's the same as `history_id` and `user_id` in this case.

bioblend/galaxy/workflows/__init__.py
nsoranzo: Because I'd like to keep the invocation methods here only for compatibility with Galaxy < 19.09, with all new features going to `InvocationClient`.

innovate-invent: Ah, ok

simonbray: Should `gi.workflows.get_invocations()` be considered deprecated then?

I still find it a handy way to filter invocations produced by a particular workflow. But I guess when this is merged it can now be done by `gi.invocations.get_invocations(workflow_id=id)` as described above.

simonbray:
Hi, is this PR still being worked on? Would be nice to be able to use these features. :)

bioblend/galaxy/workflows/__init__.py
nsoranzo: @simonbray Yep, that's the idea!

nsoranzo:
> Hi, is this PR still being worked on? Would be nice to be able to use these features. :)

Ping @innovate-invent . I'd help, but I don't have push access to your branch, I think (i.e. you disabled "Allow edits from maintainers.").

innovate-invent:
> > Hi, is this PR still being worked on? Would be nice to be able to use these features. :)
> 
> Ping @innovate-invent . I'd help, but I don't have push access to your branch, I think (i.e. you disabled "Allow edits from maintainers.").

That was selected. For some reason github has removed the option entirely :/

simonbray:
What is the status of this PR? Would it be okay if @rikeshi took it over? He is going to be spending some time over the next couple of months working on BioBlend.

innovate-invent:
You are welcome to finish it. I needed someone who knew the API better.

nsoranzo:
Replaced by https://github.com/galaxyproject/bioblend/pull/371 , thanks @innovate-invent for starting this!

PR #324, opened 2020-02-18
Fix error with toolshed get_ordered_installable_revisions and get_repository_revision_install_info

abretaud:
This fixes the following error I get when trying to get the list of revisions for a tool from the toolshed:
```
  File "/home/abretaud/miniconda3/lib/python3.7/site-packages/bioblend/toolshed/repositories/__init__.py", line 144, in get_ordered_installable_revisions
    url = self.url + '/get_ordered_installable_revisions'
AttributeError: 'ToolShedRepositoryClient' object has no attribute 'url'
```

What surprises me is that this usegalaxy.eu script should be failing without this patch, but somehow tools get updated: https://github.com/usegalaxy-eu/usegalaxy-eu-tools/blob/master/scripts/update-tool.py#L35

### DIFF
--- a/bioblend/toolshed/repositories/__init__.py
+++ b/bioblend/toolshed/repositories/__init__.py
@@ -222,7 +222,7 @@ class ToolShedRepositoryClient(Client):
                                                  'type': 'package',
                                                  'version': '0.1.18'}}]}]
         """
-        url = self.gi.url + '/repositories/get_repository_revision_install_info'
+        url = self._make_url() + '/get_repository_revision_install_info'
         params = {
             'name': name,
             'owner': owner,
--- a/bioblend/toolshed/repositories/__init__.py
+++ b/bioblend/toolshed/repositories/__init__.py
@@ -141,7 +141,7 @@ class ToolShedRepositoryClient(Client):
         :rtype: list
         :return: List of changeset revision hash strings from oldest to newest
         """
-        url = self.url + '/get_ordered_installable_revisions'
+        url = self.gi.url + '/repositories/get_ordered_installable_revisions'
         params = {
             'name': name,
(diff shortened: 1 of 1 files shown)


nsoranzo:
I think I broke this in commit 51f36db65bd0c95e9ecf73a9dd4a40bde5e5d1fd , which is not in a BioBlend release yet, which should explain why that script works.

abretaud:
Ah yes you're right, I installed from github apparently

bioblend/toolshed/repositories/__init__.py
```diff
@@ -141,7 +141,7 @@ def get_ordered_installable_revisions(self, name, owner):
         :rtype: list
         :return: List of changeset revision hash strings from oldest to newest
         """
-        url = self.url + '/get_ordered_installable_revisions'
+        url = self.gi.url + '/repositories/get_ordered_installable_revisions'
```
nsoranzo:
```suggestion
        url = self._make_url() + '/get_ordered_installable_revisions'
```

bioblend/toolshed/repositories/__init__.py
```diff
@@ -222,7 +222,7 @@ def get_repository_revision_install_info(self, name, owner,
                                                  'type': 'package',
                                                  'version': '0.1.18'}}]}]
         """
-        url = self.url + '/get_repository_revision_install_info'
+        url = self.gi.url + '/repositories/get_repository_revision_install_info'
```
nsoranzo:
```suggestion
        url = self._make_url() + '/get_repository_revision_install_info'
```

nsoranzo:
Thanks! We need to add tests for this module one day...

PR #325, opened 2020-03-06
Add get_datasets() function

simonbray:
Add function to access `/api/datasets` to get a list of all datasets, with `limit` and `offset` parameters.


bgruening:
Restarted the failing tests, looked like a network timeout.

simonbray:
One test still failing, but it doesn't seem to relate to this PR.

nsoranzo:
Definitely not related, Galaxy use of conda on macOS Python3 is broken due to https://github.com/pypa/virtualenv/issues/1704 , which is closed but not solved unfortunately.

simonbray:
In that case, can you have a look at whether this is ready to merge @nsoranzo?

nsoranzo:
Thanks @simonbray!

PR #326, opened 2020-03-25
Fix TravisCI. Drop support for Galaxy <=17.05

nsoranzo:
- Install virtualenv >=20.0.14 to fix conda commands on macOS
- Use Ubuntu bionic for TravisCI
- Deploy under Python 3.7
- Update psycopg2 requirement to a version compatible with glibc 2.26 for Galaxy releases 16.01-18.01
- Drop support for old Galaxy releases 15.03 - 17.05 which don't start without multiple patches due to broken Python eggs/wheels download: see e.g. https://travis-ci.org/github/galaxyproject/bioblend/builds/665803989 for eggs, while for wheels the psycopg2 fix above doesn't work without passing the `--extra-index-url` option to `pip`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ matrix:
             - python3.6-dev
     - stage: test
       os: osx
-      # No version of Python is available via language: on OS X workers, see https://github.com/travis-ci/travis-ci/issues/2312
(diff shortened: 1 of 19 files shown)


simonbray:
The README needs to be updated as well.

nsoranzo:
> The README needs to be updated as well.

Thanks, fixed and also removed more unnecessary code.

PR #330, opened 2020-05-07
Add job error reporting support

selten:


bioblend/galaxy/jobs/__init__.py
```diff
@@ -129,3 +131,36 @@ def get_metrics(self, job_id):
...
+        payload = {
+            "message": message,
+            "dataset_id": dataset_id
+        }
+        if email is not None:
+            payload["email"] = email
+
+        url = self._make_url(module_id=job_id) + '/error'
+        return self._post(url=url, payload=urlencode(payload, quote_via=quote_plus),
+                          headers_override={'Content-Type': 'application/x-www-form-urlencoded; charset=UTF-8'}
```
nsoranzo, line 165: I didn't have to time to check, but is it actually necessary to urlencode the payload and add the `headers_override` ? The payload is already serialised in JSON by `make_post_request()`.

selten: From my testing, this particular endpoint didn't accept 'application/json' as Content-Type.

nsoranzo: That is a problem in the Galaxy API, will be fixed by https://github.com/galaxyproject/galaxy/pull/9753

nsoranzo:
Thanks @selten! I've merged a simplified version while we wait for galaxyproject/galaxy#9753 to be merged.

PR #336, opened 2020-09-15
Do not redirect non-GET requests

simonbray:
I accidentally came across this bug while doing some work with `planemo run`.

If the URL is accidentally specified with `http://` rather than `https://`, POST requests seem not to work properly; instead, they get redirected to a GET request with the same URL. So the POST request fails, but the user is not informed, because they receive the result of a different (unintended) API call, and a status code of 200.

For example, the following:
```
http_gi = galaxy.GalaxyInstance(key='APIKEY', url='http://usegalaxy.eu/')
http_gi.histories.create_history('test')
```
will not create a history. Instead, the POST request to http://usegalaxy.eu/api/histories is redirected as a GET request, so the user just gets a list of all their histories, without any error raised.

This PR simply adds a check that the request method is 'POST' after it has been made and if not raises an error.

### DIFF
--- a/bioblend/galaxyclient.py
+++ b/bioblend/galaxyclient.py
@@ -121,6 +121,9 @@ class GalaxyClient(object):
         r = requests.post(url, data=payload, headers=headers,
                           verify=self.verify, params=post_params,
                           timeout=self.timeout)
+        if r.request.method != 'POST':
+            raise ConnectionError("POST request was unsuccessful. Check the URL is specified with https://",
+                    body='%s...' % r.text[:200], status_code=r.status_code)
         if r.status_code == 200:
             try:
                 return r.json()
--- a/bioblend/galaxyclient.py
+++ b/bioblend/galaxyclient.py
@@ -71,6 +71,7 @@ class GalaxyClient(object):
         kwargs.setdefault('verify', self.verify)
         kwargs.setdefault('timeout', self.timeout)
         r = requests.get(url, **kwargs)
+        self._check_request_method(r, 'GET')
         return r
 
     def make_post_request(self, url, payload, params=None, files_attached=False):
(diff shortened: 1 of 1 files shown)


simonbray:
Should also be extended to cover DELETE, PUT and PATCH requests.

nsoranzo:
@simonbray I believe this is probably due to a misconfiguration on usegalay.eu , in fact this doesn't happen for me if doing POST requests to http://usegalaxy.org/api/ instead of .eu

simonbray:
Strange, I encounter it with both .org and .eu.

hexylena:
@nsoranzo this is not due to misconfiguration. https://www.ietf.org/rfc/rfc2616.txt page 61-62 contain the relevant portions

>   If the 301 status code is received in response to a request other
>   than GET or HEAD, the user agent MUST NOT automatically redirect the
>   request unless it can be confirmed by the user, since this might
>   change the conditions under which the request was issued.

302 contains an identical paragraph.

simonbray:
@hexylena pointed out to me that the same effect can be achieved by using `allow_redirects=False`, thanks! Making a non-GET request then raises: `ConnectionError: Unexpected HTTP status code: 302`.

nsoranzo:
https://2.python-requests.org/en/master/user/quickstart/#redirection-and-history

nsoranzo:
See also https://github.com/psf/requests/issues/5284

bioblend/galaxyclient.py
```diff
@@ -70,7 +70,7 @@ def make_get_request(self, url, **kwargs):
         kwargs['params'] = params
         kwargs.setdefault('verify', self.verify)
         kwargs.setdefault('timeout', self.timeout)
-        r = requests.get(url, **kwargs)
+        r = requests.get(url, allow_redirects=False, **kwargs)
```
nsoranzo, line 73: I think it's fine to allow redirects for GET requests.

simonbray: Sure

nsoranzo (approved):

nsoranzo:
Thanks @simonbray and @hexylena !

Issue #339, opened 2020-10-08
[Request] Method to obtain History ID from a published History URL


mtekman:
I was pointed to this solution by @mvdbeek via @simonbray

```
https://usegalaxy.eu/api/histories/published?q=slug&qv=<history-title>
```

but I found that it was too inexact, since if a user imports and republishes the history,  the request would match multiple items.

****** 

For now I use an inelegant solution below:

```python
def get_hid_from_public_url (hist_url):
    '''Return history ID for a public Galaxy URL.'''
    try:
        req = requests.get(hist_url, timeout=3)
    except requests.ConnectionError:
        print("Failed to get contents for url")
        sys.exit(-1)

    ## Search for history id in page source
    lines = str(req.content).splitlines()
    hhid = None
    for line in lines:
        if "data: {" in line:
            hhid = line.split(' id : ')[1].split('"')[1]
            break
    return hhid
```

But it would be great if there was a `Galaxyinstance.get_history_ids(url=,name=)` function

Thanks!

mvdbeek:
slugs are unique

simonbray:
https://usegalaxy.eu/u/sbray/h/slugtest, https://usegalaxy.eu/u/simonbray/h/slugtest

(yes, I know I technically shouldn't have two accounts... Björn and Mauro are aware though)

nsoranzo:
I think that a good extension of the present API that would make this easier is to add a `published` parameter to the `HistoryClient.get_histories()` method, which could be used in combination with the `name` and `deleted` filters.

Let me know if that would be helpful and I can work on an implementation.

simonbray:
> I think that a good extension of the present API that would make this easier is to add a published parameter to the HistoryClient.get_histories() method, which could be used in combination with the name and deleted filters.

I think this would definitely be nice to have. I guess being able to filter by `slug` and `user` would also be handy for @mtekman, though it seems the latter has been disabled in the API: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/sharable.py#L459

simonbray:
I think this was fixed by https://github.com/galaxyproject/bioblend/pull/342? Please reopen if not @mtekman @nsoranzo.

PR #341, opened 2020-10-22
Add ``published`` parameter to ``HistoryClient.get_histories()``

nsoranzo:
Fix https://github.com/galaxyproject/bioblend/issues/339


simonbray:
@nsoranzo, thanks for adding this, this is very handy. However, I don't think it fixes @mtekman's issue. For that an option would also need to be added to filter on slug.

nsoranzo:
Yep, sorry, it doesn't fix it at all actually, since this let a user filter only their own published/non-published histories!

simonbray:
True - I'm afraid I didn't test this until now :(

It looks like you really need to use `/api/histories/published` then.

nsoranzo:
Indeed, we need to add a separate method to `HistoryClient`.

PR #342, opened 2020-10-30
Adding get_published_histories method

simonbray:
As discussed here: https://github.com/galaxyproject/bioblend/pull/341

Also adding the option to filter by `slug`.

### DIFF
--- a/bioblend/_tests/TestGalaxyHistories.py
+++ b/bioblend/_tests/TestGalaxyHistories.py
@@ -51,6 +51,9 @@ class TestGalaxyHistories(GalaxyTestBase.GalaxyTestBase):
         anonymous_gi = bioblend.galaxy.GalaxyInstance(url=self.gi.base_url, key=None)
         published_histories = anonymous_gi.histories.get_published_histories()
         self.assertTrue(any(h['id'] == self.history['id'] for h in published_histories))
+        history_from_slug = anonymous_gi.histories.get_published_histories(slug=updated_hist['slug'])
+        self.assertTrue(len(history_from_slug) == 1)
+        self.assertEqual(self.history['id'], history_from_slug[0]['id'])
 
     def test_get_histories(self):
         # Make sure there's at least one value - the one we created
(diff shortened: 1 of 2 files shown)


nsoranzo:
@simonbray Can you rebase this please?

bioblend/_tests/TestGalaxyHistories.py
```diff
@@ -5,6 +5,8 @@
 import tarfile
 import tempfile
 
+
```
nsoranzo, line 8:
```suggestion
```

nsoranzo:
Can you please rebase again, so we can test it under the new GitHub workflow CI?

bioblend/galaxy/histories/__init__.py
```diff
@@ -76,26 +106,45 @@ def get_histories(self, history_id=None, name=None, deleted=False, published=Non
...
-        histories = self._get(deleted=deleted, params=params)
-        if history_id is not None:
-            history = next((_ for _ in histories if _['id'] == history_id), None)
-            histories = [history] if history is not None else []
-        elif name is not None:
-            histories = [_ for _ in histories if _['name'] == name]
-        return histories
+        return self._get_histories(history_id=history_id, name=name, deleted=deleted, filter_user_published=published, get_all_published=False, slug=slug)
+
+    def get_published_histories(self, history_id=None, name=None, deleted=False, slug=None):
```
nsoranzo, line 123: I think filtering by `history_id` in a `get_*()` method is an anti-pattern, using the `show_*()` method is much more efficient, but unfortunately we have it `get_histories()` since the beginning.
I'd prefer not to add it to `get_published_histories()` though, if that's OK.

nsoranzo (approved):

nsoranzo:
@simonbray If you agree with my changes, feel free to merge!

nsoranzo:
I'm pretty sure commit d2ba08c34f5504ede86088db60d956f3d2ce6fa6 is wrong, not specifying `deleted` is equivalent to filtering out deleted histories (i.e. to `deleted=False`).

simonbray:
> I'm pretty sure commit [d2ba08c](https://github.com/galaxyproject/bioblend/commit/d2ba08c34f5504ede86088db60d956f3d2ce6fa6) is wrong, not specifying `deleted` is equivalent to filtering out deleted histories (i.e. to `deleted=False`).

You're right - thanks @nsoranzo

PR #343, opened 2020-11-02
Add rerun_job method

### LINKED ISSUE #277: Rerun error'ed dataset / job using bioblend API
I am invoking workflows using the bioblend API, but occasionally a job will fail and I need to rerun it. I currently have to log into the Galaxy instance and use the "rerun" button, remember to check the resume dependent jobs checkbox, and click submit. I'd like to automate the process. 

More than just invoking run_tool with params and inputs from the job state, I need to "replace" the job in the workflow invocation, and unpause its dependent datasets/jobs. 

How can I accomplish this using the API? 

Thanks,

- n

simonbray:
Fixes #277.

@nsoranzo, this is just a draft at the moment - if this is in principle okay for you, I'll improve the docstrings and write a test.


bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+        to get the input parameters for the job, the second to rerun it.
+
+        :type job_id: str
+        :param job_id: job ID
+
+        :rtype: dict
+        :return: Information about outputs and the rerun job
+
+        """
+        job_rerun_params = self.build_for_rerun(job_id)
```
nsoranzo, line 102:
```suggestion
        job_rerun_params = self._build_for_rerun(job_id)
```

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+        return self._get(url=url)
+
+    def remap_job(self, job_id):
+        """
+        Rerun a failed job, remapping outputs so that the failed dataset is replaced
+        with the new output.
+
+        This is a convenience bioblend method which combines two API calls - the first
+        to get the input parameters for the job, the second to rerun it.
+
```
nsoranzo, line 94: Unnecessary details IMO.

```suggestion
        If other tools were waiting for this job to finish successfully, this
        will resume those tools using the new outputs of this tool run.
```

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+        :param job_id: job ID
+
+        :rtype: dict
+        :return: A description of the given job, with all parameters required to rerun.
+
+        """
+        url = '/'.join((self._make_url(job_id), 'build_for_rerun'))
+        return self._get(url=url)
+
+    def remap_job(self, job_id):
```
nsoranzo, line 87:
```suggestion
    def rerun_and_remap_job(self, job_id):
```

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
 
         return self._get(id=job_id, params=params)
 
+    def build_for_rerun(self, job_id):
```
nsoranzo, line 73: Maybe this could be an internal method?
```suggestion
    def _build_for_rerun(self, job_id):
```

We could use this also for another `job_rerun()` method where the user can modify the tool parameters (and no output remapping).

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+        """
+        Get details of a given job of the current user.
```
nsoranzo, line 75:
```suggestion
        Get details of a given job that can be used to rerun the corresponding tool.
```

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+        :return: A description of the given job, with all parameters required to rerun.
+
+        """
+        url = '/'.join((self._make_url(job_id), 'build_for_rerun'))
+        return self._get(url=url)
+
+    def remap_job(self, job_id):
+        """
+        Rerun a failed job, remapping outputs so that the failed dataset is replaced
+        with the new output.
```
nsoranzo, line 90:
```suggestion
        Rerun a failed job, remapping outputs so that the failed datasets are
        replaced with the new outputs.
```

nsoranzo:
Fantastic, thanks a lot!

simonbray:
Thanks for the comments @nsoranzo!

@bgruening also wanted to include a `gi.histories.rerun_failed_jobs()` method to rerun all failed jobs in a history, but I was a bit sceptical as this would entail a very large number of API calls within a single method, which I don't see anywhere else in bioblend. Is it a deliberate aim of bioblend that each method corresponds to a single API endpoint / call, or is this just coincidence?

Or maybe the solution is to implement this upstream as a new API endpoint?

bioblend/galaxy/jobs/__init__.py
simonbray: Or maybe a single `job_rerun()` method with a boolean variable `remap`?

nsoranzo: That could work, but that we would also need to add a `tool_inputs_update` dict parameter to let the user update tool parameters.

nsoranzo:
> @bgruening also wanted to include a `gi.histories.rerun_failed_jobs()` method to rerun all failed jobs in a history, but I was a bit sceptical as this would entail a very large number of API calls within a single method, which I don't see anywhere else in bioblend.

I'd love to have such a method! We already have other methods performing multiple API calls, e.g. `HistoryClient.export_history()` and `DatasetClient.download_dataset()` .

> Is it a deliberate aim of bioblend that each method corresponds to a single API endpoint / call, or is this just coincidence?

I don't think that's BioBlend's aim, our `ABOUT.rst` says: "BioBlend’s goal is to make it easier to script and automate the running of Galaxy analyses,..."

simonbray:
I'm just testing this out and it appears there are two different kinds of job ids - one with 32 chars and one with 16 (the latter are older datasets). It seems this PR doesn't work with the old version - the post request succeeds and returns the expected dictionary, but nothing visible changes in the UI.

Moreover when I look closely at the dictionary which is created as output, the history ID is different to the original one, and I can't find the new history anywhere in the UI! Really strange. :/

nsoranzo:
@simonbray It seems there is some Galaxy bug lurking there.

simonbray:
Okay, I spent some time thinking about writing a test, but this doesn't seem very straightforward - a workflow, or a tool with a collection as input, has to be run in such a way that it's guaranteed to fail, so it can be rerun. Anyway, there are no other test cases for the job client currently.

Any other thoughts @nsoranzo? Otherwise I think this can be merged.

bioblend/galaxy/jobs/__init__.py
simonbray: I think I prefer to keep this as a normal (non-internal) method, isn't it possible that a user might want to use it directly?

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+
+        If other tools were waiting for this job to finish successfully, this
+        will resume those tools using the new outputs of this tool run.
+        :type job_id: str
+        :param job_id: job ID
+
+        :rtype: dict
+        :return: Information about outputs and the rerun job
+        .. note::
+          Only possible for Galaxy 20.09 and later.
```
nsoranzo, line 100:
```suggestion

        .. note::
          This method works only for Galaxy ``release_20.09`` and later.
```

bioblend/galaxy/jobs/__init__.py
nsoranzo: > I think I prefer to keep this as a normal (non-internal) method, isn't it possible that a user might want to use it directly?

What would be the use case?

mvdbeek:
Can you list the history contents via the API ?

> Okay, I spent some time thinking about writing a test, but this doesn't seem very straightforward - a workflow, or a tool with a collection as input, has to be run in such a way that it's guaranteed to fail, so it can be rerun.

Not suggesting you need to do that, but you can start galaxy with `GALAXY_RUN_WITH_TEST_TOOLS=1 sh run.sh` and you'll get some framework testing tools. https://github.com/mvdbeek/galaxy/blob/f8d3737ecf3a162920ca84c8edc3704e25c49d12/test/functional/tools/job_properties.xml#L26 let's you fail a tool using a parameter.

simonbray:
> > I think I prefer to keep this as a normal (non-internal) method, isn't it possible that a user might want to use it directly?
>
> What would be the use case?

The user might want to get the parameters and then modify them prior to rerunning.

simonbray:
> > Okay, I spent some time thinking about writing a test, but this doesn't seem very straightforward - a workflow, or a tool with a collection as input, has to be run in such a way that it's guaranteed to fail, so it can be rerun.
> 
> Not suggesting you need to do that, but you can start galaxy with `GALAXY_RUN_WITH_TEST_TOOLS=1 sh run.sh` and you'll get some framework testing tools. https://github.com/mvdbeek/galaxy/blob/f8d3737ecf3a162920ca84c8edc3704e25c49d12/test/functional/tools/job_properties.xml#L26 let's you fail a tool using a parameter.

Interesting, thanks - I'll have a look.

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,48 @@ def show_job(self, job_id, full_details=False):
...
+        :param job_id: job ID
+
+        :rtype: dict
+        :return: A description of the given job, with all parameters required to rerun.
+
+        """
+        url = '/'.join((self._make_url(job_id), 'build_for_rerun'))
+        return self._get(url=url)
+
+    def rerun_and_remap_job(self, job_id):
```
nsoranzo, line 87: Possible signature for a general `rerun_job()` method as discussed in https://github.com/galaxyproject/bioblend/pull/343/files#r516030456 :

```suggestion
    def rerun_job(self, job_id, remap=False, tool_inputs_update=None):
```

nsoranzo:
> > > I think I prefer to keep this as a normal (non-internal) method, isn't it possible that a user might want to use it directly?
> > 
> > 
> > What would be the use case?
> 
> The user might want to get the parameters and then modify them prior to rerunning.

Yes, but I'd still keep `build_for_rerun` as an internal method for the more general `rerun_job()` we were discussing, see https://github.com/galaxyproject/bioblend/pull/343#discussion_r518222561 for its possible signature.

simonbray:
Okay, I turned this into a general `rerun_job` method and wrote a test for the simple case (i.e. no remapping). Unfortunately this is failing, which seems to reveal a bug in the `build_for_rerun` API. I'm using the random_lines tool for the test and specifying the conditional parameters with `'seed_source|seed_source_selector': 'set_seed', 'seed_source|seed': 'asdf'`. However, `build_for_rerun` gives `'seed_source': {'seed_source_selector': 'set_seed', 'seed': 'asdf'}` and the latter doesn't set the tool params correctly.

bioblend/_tests/TestGalaxyJobs.py
```diff
@@ -0,0 +1,26 @@
+from . import (
+    GalaxyTestBase,
+    test_util
+)
+
+
+class TestGalaxyJobs(GalaxyTestBase.GalaxyTestBase):
+
+    @test_util.skip_unless_tool("random_lines1")
```
nsoranzo, line 9:
```suggestion
    @test_util.skip_unless_galaxy('release_20.09')
    @test_util.skip_unless_tool("random_lines1")
```

simonbray: I think a normal rerun (which is what is being tested here) should be fine with older versions, right? It's only the remapping which won't work.

mvdbeek: The format of the state_inputs changed with 20.09, I don't think that works with older Galaxy versions.

simonbray: Okay, good to know.

nsoranzo:
> Okay, I turned this into a general `rerun_job` method and wrote a test for the simple case (i.e. no remapping). Unfortunately this is failing, which seems to reveal a bug in the `build_for_rerun` API. I'm using the random_lines tool for the test and specifying the conditional parameters with `'seed_source|seed_source_selector': 'set_seed', 'seed_source|seed': 'asdf'`. However, `build_for_rerun` gives `'seed_source': {'seed_source_selector': 'set_seed', 'seed': 'asdf'}` and the latter doesn't set the tool params correctly.

I bet that if you use `self.gi.tools.run_tool()` instead of making directly the POST request it will work (the secret being the `payload["inputs"] = tool_inputs.to_dict()` line).

Also, can you maybe use a different tool in test? With `random_lines1` the test fails, well, randomly!

simonbray:
> I bet that if you use self.gi.tools.run_tool() instead of making directly the POST request it will work (the secret being the payload["inputs"] = tool_inputs.to_dict() line).

This line is for the case when the inputs are submitted as an object of the `InputsBuilder` class, right, like in the tools tests? We are getting the inputs from `build_for_rerun` as a dictionary, so `payload["inputs"] = tool_inputs.to_dict()` will just give `AttributeError: 'dict' object has no attribute 'to_dict'`.

Unless you are proposing constructing this inputs object from the dictionary? But this seems complicated, it would be nice if the dictionary just worked.

> Also, can you maybe use a different tool in test? With random_lines1 the test fails, well, randomly!

Yes, makes sense :smile:

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,66 @@ def show_job(self, job_id, full_details=False):
...
+        :rtype: dict
+        :return: Information about outputs and the rerun job
+        .. note::
+          This method can only be used with Galaxy ``release_20.09`` or later.
+        """
+        job_rerun_params = self._build_for_rerun(job_id)
+        job_inputs = job_rerun_params['state_inputs']
+
+        if remap:
+            assert job_inputs['job_remap'], 'remap was set to True, but this job is not remappable.'
```
nsoranzo, line 118:
```suggestion
            if not job_inputs['job_remap']:
                raise ValueError('remap was set to True, but this job is not remappable.')
```

bioblend/galaxy/tools/__init__.py
```diff
@@ -134,7 +134,7 @@ def show_tool(self, tool_id, io_details=False, link_details=False):
         params['link_details'] = link_details
         return self._get(id=tool_id, params=params)
 
-    def run_tool(self, history_id, tool_id, tool_inputs):
+    def run_tool(self, history_id, tool_id, tool_inputs, input_format='21.01'):
```
nsoranzo, line 137:
```suggestion
    def run_tool(self, history_id, tool_id, tool_inputs, input_format='legacy'):
```

bioblend/_tests/TestGalaxyJobs.py
```diff
@@ -0,0 +1,27 @@
...
+        dataset_id = self._test_dataset(history_id, contents="line 1\nline 2\rline 3\r\nline 4")
+        tool_inputs = {'num_lines': '1',
+                       'input': {'src': 'hda', 'id': dataset_id},
+                       'seed_source': {'seed_source_selector': 'set_seed',
+                                       'seed': 'asdf'}
+                       }
+        original_output = self.gi.tools.run_tool(
+            history_id=history_id,
+            tool_id="random_lines1",
+            tool_inputs=tool_inputs
```
nsoranzo, line 21:
```suggestion
            tool_inputs=tool_inputs,
            input_format='21.01'
```

nsoranzo:
Can you please rebase, so we can test it under the new GitHub workflow CI?

bioblend/galaxy/jobs/__init__.py
```diff
@@ -70,6 +70,70 @@ def show_job(self, job_id, full_details=False):
...
+        url = '/'.join((self.gi.url, 'tools'))
+        payload = {
+            "history_id": history_id if history_id else job_rerun_params['history_id'],
+            "tool_id": job_rerun_params['id'],
+            "inputs": job_inputs
+
+        }
+        params = {
+            "input_format": '21.01'
+        }
```
nsoranzo, line 134: I think you need this here (instead of inside the `payload` dict) because of https://github.com/galaxyproject/galaxy/pull/10720/files#diff-f12168394d9b932c4e06147dc6b4fda47b6947db99d7fab633789d62815bd009R539 which should have been `input_format = payload.get('input_format', 'legacy')`

Can we fix Galaxy 21.01 instead?

simonbray: Indeed - are there some guidelines for which information should be conveyed via `params` and which via `payload`?

simonbray: I created a PR here: https://github.com/galaxyproject/galaxy/pull/11332

nsoranzo: GET requests use `params`, POST and PUT requests use `payload`

simonbray: That makes sense to me, but then why do many of the methods defining POST API endpoints (e.g. that `create` one) take `**kwds` arguments?

nsoranzo: It may be due to how routing is implemented, but I'm not very sure to be honest.

nsoranzo (approved):
Huge thanks @simonbray for implementing this fundamental feature!!

simonbray:
Phew, finally. Thanks @nsoranzo and @mvdbeek for your help!

PR #368, opened 2021-02-12
add InvocationClient methods

rikeshi:
This PR adds InvocationClient methods for galaxy API endpoints that were missing.


bioblend/galaxy/invocations/__init__.py
```diff
@@ -181,6 +212,32 @@ def get_invocation_report(self, invocation_id):
         url = self._make_url(invocation_id) + '/report'
         return self._get(url=url)
 
+    def get_invocation_report_pdf(self, invocation_id):
```
simonbray, line 215: Can you combine this with the method above by including a `format`argument which defaults to markdown but can also be set to pdf?

The PDF also only renders if the weasyprint dependency is installed, see: https://github.com/galaxyproject/galaxy/pull/9651. Can you raise an error if this is the case?

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -34,6 +34,21 @@ def test_invocation(self):
...
+            inputs_by='name',
+        )
+
+        time.sleep(5)
+        step_jobs_summary = self.gi.invocations.get_invocation_step_jobs_summary(invocation2['id'])
+        self.assertEqual(len(step_jobs_summary), 1)
+        self.assertEqual(step_jobs_summary[0]['populated_state'], 'ok')
+
+        biocompute_object = self.gi.invocations.get_invocation_biocompute_object(invocation2['id'])
+        self.assertEqual(len(biocompute_object['description_domain']['pipeline_steps']), 1)
```
nsoranzo, line 50: The biocompute object export only works on Galaxy >=20.09, can you split it in a separate test annotated with `@test_util.skip_unless_galaxy('release_20.09')`?

It probably makes sense to also makes another test for lines 37-47 (and rename the original test `test_cancel_workflow`), with a `_invoke_workflow()` setup method containing lines 10-21 to be re-used by all 3 tests.

bioblend/galaxy/invocations/__init__.py
```diff
@@ -178,7 +212,27 @@ def get_invocation_report(self, invocation_id):
...
-        url = self._make_url(invocation_id) + '/report'
+        if format == 'markdown':
+            url = self._make_url(invocation_id) + '/report'
+            return self._get(url=url)
+        elif format == 'pdf':
+            url = self._make_url(invocation_id) + '/report.pdf'
+            try:
+                return self._get(url=url, json=False)
+            except Exception as e:
+                raise e
```
nsoranzo, line 223:
```suggestion
                raise Exception(f"Failed to get the PDF report, the necessary dependencies may not be installed on the Galaxy server: {e}")
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -160,14 +160,48 @@ def get_invocation_summary(self, invocation_id):
...
+                {'populated_state': 'ok',
+                 'states': {'new': 1},
+                 'model': 'Job',
+                 'id': '2a56795cad3c7db3'}
+            ]
+        """
+        url = self._make_url(invocation_id) + '/step_jobs_summary'
+        return self._get(url=url)
+
+    def get_invocation_report(self, invocation_id, format='markdown'):
```
nsoranzo, line 194: Should we require an output file object as argument to save the PDF report, similarly to https://github.com/galaxyproject/bioblend/blob/main/bioblend/galaxy/histories/__init__.py#L652-L679 ?
In that case, maybe keeping the PDF report method separate could be better @simonbray ?

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -29,10 +18,34 @@ def test_invocation(self):
...
+
+    def test_get_invocation_report(self):
+        invocation = self._invoke_workflow()
+
+        invocation_id = invocation['id']
+        workflow_id = invocation['workflow_id']
         report = self.gi.invocations.get_invocation_report(invocation_id)
+        assert report['workflows'] == {workflow_id: {'name': 'paste_columns'}}
+        with self.assertRaises(Exception):
+            self.gi.invocations.get_invocation_report(invocation_id, format='pdf')
```
nsoranzo, line 31:
```suggestion
        try:
            self.gi.invocations.get_invocation_report(invocation_id, format='pdf')
        except Exception:
            # This can fail if dependencies as weasyprint are not installed on the Galaxy server
            pass
```

bioblend/galaxy/invocations/__init__.py
simonbray: Yeah, I just realised that the PDF should be saved as a output file... sorry @rikeshi, can you revert back to the old version with a separate method, and include a `file_path` argument?

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -29,10 +18,33 @@ def test_invocation(self):
...
 
-        assert summary['states'] == {}
-        assert report['workflows'] == {workflow['id']: {'name': 'paste_columns'}}
+    @test_util.skip_unless_galaxy('release_20.09')
+    def test_get_invocation_biocompute_object(self):
+        invocation = self._invoke_workflow()
+
+        time.sleep(3)
```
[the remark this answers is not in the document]
rikeshi: Okay, I will do that.

bioblend/galaxy/invocations/__init__.py
rikeshi: Okay

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -70,6 +79,32 @@ def invocation_steps_by_order_index():
...
+        def invocation_steps_by_order_index():
+            invocation = self.gi.invocations.show_invocation(invocation_id)
+            return {s["order_index"]: s for s in invocation["steps"]}
+
+        for _ in range(20):
+            if 2 in invocation_steps_by_order_index():
+                break
+            time.sleep(.5)
+
+        return invocation_steps_by_order_index()
```
nsoranzo, line 110:
```suggestion
        for _ in range(20):
            invocation = self.gi.invocations.show_invocation(invocation_id)
            if invocation["state"] == "scheduled":
                break
            time.sleep(.5)
        else:
            invocation = self.gi.invocations.show_invocation(invocation_id)
            self.assertEqual(invocation["state"], "scheduled")
```

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -49,16 +66,8 @@ def test_workflow_scheduling(self):
...
 
-        def invocation_steps_by_order_index():
-            invocation = self.gi.invocations.show_invocation(invocation_id)
-            return {s["order_index"]: s for s in invocation["steps"]}
-
-        for _ in range(20):
-            if 2 in invocation_steps_by_order_index():
-                break
-            time.sleep(.5)
+        steps = self._wait_invocation(invocation_id)
```
nsoranzo, line 70: @rikeshi I think you moved the wrong `for` loop to `_wait_invocation()`:

```suggestion
        def invocation_steps_by_order_index():
            invocation = self.gi.invocations.show_invocation(invocation_id)
            return {s["order_index"]: s for s in invocation["steps"]}

        for _ in range(20):
            if 2 in invocation_steps_by_order_index():
                break
            time.sleep(.5)
        steps = invocation_steps_by_order_index()
```

rikeshi: That was the only for loop, or not? The removed line and added lines in your suggestion are equivalent since `_wait_invocation()` wraps all of it and returns the steps at the end. Maybe what I did is not what you had in mind. Could you clarify?

rikeshi: Oh, I see now. I will fix it!

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -70,6 +79,32 @@ def invocation_steps_by_order_index():
                 break
 
             time.sleep(.5)
+        else:
+            invocation = self.gi.invocations.show_invocation(invocation_id)
+            self.assertEqual(invocation["state"], "scheduled")
```
nsoranzo, line 84: This is the `for` loop that should be replaced with:
```python
        self._wait_invocation(invocation_id)
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -160,13 +161,47 @@ def get_invocation_summary(self, invocation_id):
...
         """
         Get a Markdown report for an invocation.
 
         :type invocation_id: str
         :param invocation_id: Encoded workflow invocation ID
 
+        :type format: str
+        :param format: The format of the report, can be 'markdown' or 'pdf'
+
```
nsoranzo, line 204:
```suggestion
```

bioblend/galaxy/invocations/__init__.py
nsoranzo:
```suggestion
    def get_invocation_report(self, invocation_id):
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -181,6 +216,42 @@ def get_invocation_report(self, invocation_id):
...
+        Get a PDF report for an invocation.
+
+        :type invocation_id: str
+        :param invocation_id: Encoded workflow invocation ID
+
+        :type file_path: str
+        :param file_path: Path to save the file under
+
+        :rtype: bytes
+        :return: The invocation report.
```
nsoranzo, line 230:
```suggestion
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -181,6 +216,42 @@ def get_invocation_report(self, invocation_id):
...
+        """
+        url = self._make_url(invocation_id) + '/report.pdf'
+        try:
+            r = self.gi.make_get_request(url, stream=True, json=False)
```
nsoranzo, line 234: I don't think `json=Falsee` is needed?

```suggestion
            r = self.gi.make_get_request(url, stream=True)
```

rikeshi: Okay

bioblend/galaxy/invocations/__init__.py
rikeshi: Good catch, missed that

nsoranzo:
LGTM, thanks @rikeshi !

@simonbray ?

bioblend/galaxy/invocations/__init__.py
```diff
@@ -181,6 +213,39 @@ def get_invocation_report(self, invocation_id):
...
+        """
+        url = self._make_url(invocation_id) + '/report.pdf'
+        try:
+            r = self.gi.make_get_request(url, stream=True)
+            r.raise_for_status()
+            with open(file_path, 'wb') as outf:
+                for chunk in r.iter_content(chunk_size):
+                    outf.write(chunk)
+        except Exception as e:
+            raise Exception(f"Failed to get the PDF report, the necessary dependencies may not be installed on the Galaxy server: {e}")
```
simonbray, line 234:
```suggestion
        r = self.gi.make_get_request(url, stream=True)
        if not r.ok:
            raise Exception(f"Failed to get the PDF report, the necessary dependencies may not be installed on the Galaxy server: {e}")
        with open(file_path, 'wb') as outf:
            for chunk in r.iter_content(chunk_size):
                outf.write(chunk)
           
```

simonbray: I think this is neater. Or is there a reason writing the file needs to be in the try-except block?

simonbray (commented):
Looks pretty good to me

bioblend/galaxy/invocations/__init__.py
```diff
@@ -160,6 +161,37 @@ def get_invocation_summary(self, invocation_id):
...
 
+    def get_invocation_step_jobs_summary(self, invocation_id):
+        """
+        Get a summary of an invocation, stating the number of jobs which
+        succeed, which are paused and which have errored.
+
+        :type invocation_id: str
+        :param invocation_id: Encoded workflow invocation ID
+
+        :rtype: dict[]
```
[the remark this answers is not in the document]
rikeshi: Doesn't this method return a list of dictionaries?

bioblend/galaxy/invocations/__init__.py
rikeshi: No, I assumed it was needed because of the get request that could fail. This is definitely neater.

bioblend/galaxy/invocations/__init__.py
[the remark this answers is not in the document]
simonbray: Yes, you are right. Is that what `dict[]` is meant to represent?

In the documentation so far it looks like we just wrote `list of dicts` for the rtype

simonbray (approved):

bioblend/galaxy/invocations/__init__.py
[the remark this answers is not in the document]
rikeshi: Ok, changed it

nsoranzo:
Thanks @rikeshi !

PR #369, opened 2021-02-15
WorkflowsClient methods

rikeshi:
add

`
WorkflowsClient.show_versions()
WorkflowsClient.create_workflow_from_history()
WorkflowsClient.refactor()
`

as discussed with @simonbray

### DIFF
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -1,10 +1,10 @@
 import os
+import time
 
 import bioblend
 import bioblend.galaxy
 from . import test_util
 from .test_util import unittest
-import time
 
 bioblend.set_stream_logger('test', level='INFO')
 
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -27,17 +27,22 @@ class GalaxyTestBase(unittest.TestCase):
         dataset_contents = self.gi.datasets.download_dataset(dataset_id, maxwait=timeout_seconds)
         self.assertEqual(dataset_contents, expected_contents)
 
-    def _wait_invocation(self, invocation_id, wait_steps=False):
+    def _wait_invocation(self, invocation, wait_steps=False):
+        galaxy_version = os.environ.get('GALAXY_VERSION', None)
+        is_newer = galaxy_version is None or galaxy_version >= 'release_20.01'
+        show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
+        args = [invocation['id']] if is_newer else [invocation['id'], invocation['workflow_id']]
         for _ in range(20):
-            invocation = self.gi.invocations.show_invocation(invocation_id)
+            invocation = show_invocation(*args)
             if invocation["state"] == "scheduled":
                 if wait_steps:
-                    for step in invocation["steps"]:
-                        if step["state"] == "scheduled":
-                            break
+                    if "steps" in invocation:
+                        for step in invocation["steps"]:
+                            if step["state"] == "scheduled":
+                                break
                 else:
                     break
             time.sleep(.5)
         else:
-            invocation = self.gi.invocations.show_invocation(invocation_id)
+            invocation = show_invocation(*args)
             self.assertEqual(invocation["state"], "scheduled")
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -31,7 +31,7 @@ class GalaxyTestBase(unittest.TestCase):
         galaxy_version = os.environ.get('GALAXY_VERSION', None)
         is_newer = galaxy_version is None or galaxy_version >= 'release_19.09'
         show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
-        args = [invocation['id']] if is_newer else [invocation['id'], invocation['workflow_id']]
+        args = [invocation['id']] if is_newer else [invocation['workflow_id'], invocation['id']]
         for _ in range(20):
             invocation = show_invocation(*args)
             if invocation["state"] == "scheduled":
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -4,6 +4,7 @@ import bioblend
 import bioblend.galaxy
 from . import test_util
 from .test_util import unittest
+import time
 
 bioblend.set_stream_logger('test', level='INFO')
 
@@ -25,3 +26,13 @@ class GalaxyTestBase(unittest.TestCase):
     def _wait_and_verify_dataset(self, dataset_id, expected_contents, timeout_seconds=BIOBLEND_TEST_JOB_TIMEOUT):
         dataset_contents = self.gi.datasets.download_dataset(dataset_id, maxwait=timeout_seconds)
         self.assertEqual(dataset_contents, expected_contents)
+
+    def _wait_invocation(self, invocation_id):
+        for _ in range(20):
+            invocation = self.gi.invocations.show_invocation(invocation_id)
+            if invocation["state"] == "scheduled":
+                break
+            time.sleep(.5)
+        else:
+            invocation = self.gi.invocations.show_invocation(invocation_id)
+            self.assertEqual(invocation["state"], "scheduled")
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -29,7 +29,7 @@ class GalaxyTestBase(unittest.TestCase):
 
     def _wait_invocation(self, invocation, wait_steps=False):
         galaxy_version = os.environ.get('GALAXY_VERSION', None)
-        is_newer = galaxy_version is None or galaxy_version >= 'release_20.01'
+        is_newer = galaxy_version is None or galaxy_version >= 'release_19.09'
         show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
         args = [invocation['id']] if is_newer else [invocation['id'], invocation['workflow_id']]
         for _ in range(20):
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -27,19 +27,23 @@ class GalaxyTestBase(unittest.TestCase):
         dataset_contents = self.gi.datasets.download_dataset(dataset_id, maxwait=timeout_seconds)
         self.assertEqual(dataset_contents, expected_contents)
 
-    def _wait_invocation(self, invocation, wait_steps=False):
+    def _wait_invocation(self, invocation, wait_steps=False, timeout_seconds=10):
         galaxy_version = os.environ.get('GALAXY_VERSION', None)
         is_newer = galaxy_version == 'dev' or galaxy_version >= 'release_19.09'
         show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
         args = [invocation['id']] if is_newer else [invocation['workflow_id'], invocation['id']]
-        for _ in range(20):
+        for _ in range(timeout_seconds * 2):
             invocation = show_invocation(*args)
             if invocation["state"] == "scheduled":
                 if wait_steps:
                     if "steps" in invocation:
+                        steps_scheduled = True
                         for step in invocation["steps"]:
-                            if step["state"] == "scheduled":
+                            if step["state"] != "scheduled":
+                                steps_scheduled = False
                                 break
+                        if steps_scheduled:
+                            break
                 else:
                     break
             time.sleep(.5)
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -29,7 +29,7 @@ class GalaxyTestBase(unittest.TestCase):
 
     def _wait_invocation(self, invocation, wait_steps=False):
         galaxy_version = os.environ.get('GALAXY_VERSION', None)
-        is_newer = galaxy_version is None or galaxy_version >= 'release_19.09'
+        is_newer = galaxy_version == 'dev' or galaxy_version >= 'release_19.09'
         show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
         args = [invocation['id']] if is_newer else [invocation['workflow_id'], invocation['id']]
         for _ in range(20):
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -1,5 +1,4 @@
 import os
-import time
 
 import bioblend
 import bioblend.galaxy
@@ -26,27 +25,3 @@ class GalaxyTestBase(unittest.TestCase):
     def _wait_and_verify_dataset(self, dataset_id, expected_contents, timeout_seconds=BIOBLEND_TEST_JOB_TIMEOUT):
         dataset_contents = self.gi.datasets.download_dataset(dataset_id, maxwait=timeout_seconds)
         self.assertEqual(dataset_contents, expected_contents)
-
-    def _wait_invocation(self, invocation, wait_steps=False, timeout_seconds=10):
-        galaxy_version = os.environ.get('GALAXY_VERSION', None)
-        is_newer = galaxy_version == 'dev' or galaxy_version >= 'release_19.09'
-        show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
-        args = [invocation['id']] if is_newer else [invocation['workflow_id'], invocation['id']]
-        for _ in range(timeout_seconds * 2):
-            invocation = show_invocation(*args)
-            if invocation["state"] == "scheduled":
-                if wait_steps:
-                    if "steps" in invocation:
-                        steps_scheduled = True
-                        for step in invocation["steps"]:
-                            if step["state"] != "scheduled":
-                                steps_scheduled = False
-                                break
-                        if steps_scheduled:
-                            break
-                else:
-                    break
-            time.sleep(.5)
-        else:
-            invocation = show_invocation(*args)
-            self.assertEqual(invocation["state"], "scheduled")
--- a/bioblend/_tests/GalaxyTestBase.py
+++ b/bioblend/_tests/GalaxyTestBase.py
@@ -27,11 +27,16 @@ class GalaxyTestBase(unittest.TestCase):
         dataset_contents = self.gi.datasets.download_dataset(dataset_id, maxwait=timeout_seconds)
         self.assertEqual(dataset_contents, expected_contents)
 
-    def _wait_invocation(self, invocation_id):
+    def _wait_invocation(self, invocation_id, wait_steps=False):
         for _ in range(20):
             invocation = self.gi.invocations.show_invocation(invocation_id)
             if invocation["state"] == "scheduled":
-                break
+                if wait_steps:
+                    for step in invocation["steps"]:
+                        if step["state"] == "scheduled":
+                            break
+                else:
+                    break
             time.sleep(.5)
         else:
             invocation = self.gi.invocations.show_invocation(invocation_id)
--- a/bioblend/_tests/TestGalaxyWorkflows.py
+++ b/bioblend/_tests/TestGalaxyWorkflows.py
@@ -209,12 +209,32 @@ class TestGalaxyWorkflows(GalaxyTestBase.GalaxyTestBase):
         assert invoke_response['state'] == 'new', invoke_response
 
     def test_create_workflow_from_history(self):
-        invoke_response = self._invoke_workflow()
-        wf_id = invoke_response['workflow_id']
-        history_id = invoke_response['history_id']
-        self.gi.workflows.cancel_invocation(wf_id, invoke_response['id'])
-        wf2 = self.gi.workflows.create_workflow_from_history(history_id, 'My new workflow!')
+        path = test_util.get_abspath(os.path.join('data', 'paste_columns.ga'))
+        wf1 = self.gi.workflows.import_workflow_from_local_path(path)
+        history_id = self.gi.histories.create_history(name="test_wf")['id']
+        dataset_id = self._test_dataset(history_id)
+        dataset = {'src': 'hda', 'id': dataset_id}
+        invoc1 = self.gi.workflows.invoke_workflow(
+            wf1['id'],
+            inputs={'Input 1': dataset, 'Input 2': dataset},
+            history_id=history_id,
+            inputs_by='name',
+        )
+        self._wait_invocation(invoc1['id'])
+        wf1 = self.gi.workflows.show_workflow(wf1['id'])
+        history_id = invoc1['history_id']
+        datasets = self.gi.datasets.get_datasets()
+        dataset_ids = [dataset['id'] for dataset in datasets]
+        # jobs = self.gi.jobs.get_jobs()
+        # job_ids = [job['id'] for job in jobs if job['history_id'] == history_id]
+        wf2 = self.gi.workflows.create_workflow_from_history(history_id, 'My new workflow!',
+                                                             dataset_ids=dataset_ids)
+        # , job_ids=job_ids)
+        wf2 = self.gi.workflows.show_workflow(wf2['id'])
         self.assertEqual(wf2['name'], 'My new workflow!')
+        self.assertTrue('steps' in wf1)
+        self.assertTrue('steps' in wf2)
+        self.assertEqual(len(wf1['steps']), len(wf2['steps']))
 
     def test_show_versions(self):
         invoke_response = self._invoke_workflow()
--- a/bioblend/_tests/TestGalaxyWorkflows.py
+++ b/bioblend/_tests/TestGalaxyWorkflows.py
@@ -237,7 +237,6 @@ class TestGalaxyWorkflows(GalaxyTestBase.GalaxyTestBase):
     def test_show_versions(self):
         invoke_response = self._invoke_workflow()
         wf_id = invoke_response['workflow_id']
-        self.gi.workflows.cancel_invocation(wf_id, invoke_response['id'])
         versions = self.gi.workflows.show_versions(wf_id)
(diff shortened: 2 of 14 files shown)


bioblend/galaxy/workflows/__init__.py
```diff
@@ -730,6 +730,80 @@ def delete_workflow(self, workflow_id):
...
+            "job_ids": job_ids if job_ids else [],
+            "dataset_ids": dataset_ids if job_ids else [],
+            "dataset_collection_ids": dataset_collection_ids if dataset_collection_ids else [],
+            "workflow_name": workflow_name
+        }
+        return self._post(payload=payload)
+
+    def show_versions(self, workflow_id):
+        """
+        Get versions for a workflow_id.
```
nsoranzo, line 796:
```suggestion
        Get versions for a workflow.
```

rikeshi:
The test for show_versions should fail because of a error in the Galaxy API method. I created a PR for that here: https://github.com/galaxyproject/galaxy/pull/11373

nsoranzo:
> The test for show_versions should fail because of a error in the Galaxy API method. I created a PR for that here: [galaxyproject/galaxy#11373](https://github.com/galaxyproject/galaxy/pull/11373)

Thanks for letting us know, we can restart the test when that's merged.

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,64 @@ def test_run_workflow(self):
...
+        invoke_response = self._invoke_workflow()
+        assert invoke_response['state'] == 'new', invoke_response
+
+    def test_create_workflow_from_history(self):
+        invoke_response = self._invoke_workflow()
+        wf_id = invoke_response['workflow_id']
+        history_id = invoke_response['history_id']
+        self.gi.workflows.cancel_invocation(wf_id, invoke_response['id'])
+        wf2 = self.gi.workflows.create_workflow_from_history(history_id, 'My new workflow!')
+        self.assertEqual(wf2['name'], 'My new workflow!')
```
simonbray, line 217: Can you make the test a bit tighter here? E.g. by comparing the number of steps to the original workflow.

Btw, I think you need to submit the `job_ids` and other parameters, otherwise you get an empty workflow.

rikeshi: Sorry, I cannot get it to work. Passing job_ids gives me "Attempt to create workflow with job not connected to current history". Passing dataset_ids results in a workflow with no steps.

simonbray: Can you try something like `job_ids = [job['id'] for job in self.gi.jobs.get_jobs() if job['history_id'] == history_id]`?

For the datasets, you should pass the hid, not the id, i.e. the index of the dataset in the history. So `[1, 2]` should be enough.

rikeshi: I tried it but it makes no difference to the output.

simonbray: After a bit of fiddling, it's once again a time issue - the invocation needs to fully complete before the job_ids can be collected (see the latest commit).The `Attempt to create workflow with job not connected to current history` is misleading.

Can you ensure _wait_invocation waits until all individual steps are scheduled, not just the invocation itself? That should hopefully fix it.

simonbray: Another point: we should not duplicate the _wait_invocation method. Can you move this to GalaxyTestBase.py to reduce boilerplate?

rikeshi:
Job summaries pre-19.05 do not seem to include 'history_id'. Is there another way to get jobs associated with a history? Or should the test be skipped for those versions?

simonbray:
That's annoying, try something like
```
datasets = gi.histories.show_history(history_id, contents=True)
jobs = [gi.datasets.show_dataset(dataset['id'])['creating_job'] for dataset in datasets]
```

bioblend/_tests/GalaxyTestBase.py
```diff
@@ -25,3 +26,23 @@ def _test_dataset(self, history_id, contents="1\t2\t3", **kwds):
     def _wait_and_verify_dataset(self, dataset_id, expected_contents, timeout_seconds=BIOBLEND_TEST_JOB_TIMEOUT):
         dataset_contents = self.gi.datasets.download_dataset(dataset_id, maxwait=timeout_seconds)
         self.assertEqual(dataset_contents, expected_contents)
+
+    def _wait_invocation(self, invocation, wait_steps=False):
```
nsoranzo, line 30: Should we maybe move and expose this as a convenience `wait_for_invocation()` method of `InvocationClient` ?

rikeshi: Instead of asserting that it was scheduled, would it then be better to return `True` if the scheduling was successful, and `False` if time ran out?

bioblend/_tests/TestGalaxyWorkflows.py
nsoranzo: > After a bit of fiddling, it's once again a time issue - the invocation needs to fully complete before the job_ids can be collected (see the latest commit).The `Attempt to create workflow with job not connected to current history` is misleading.
> 
> Can you ensure _wait_invocation waits until all individual steps are scheduled, not just the invocation itself? That should hopefully fix it.

@simonbray From my reading of Galaxy's `lib/galaxy/workflow/run.py` an invocation state should be set to `scheduled` only when all the steps have also been scheduled (though the jobs may still have to be).

bioblend/_tests/GalaxyTestBase.py
nsoranzo: In other parts of BioBlend we raise a specific exception (e.g. `DatasetTimeoutException`), like [requests](https://requests.readthedocs.io/) does.

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,89 @@ def delete_workflow(self, workflow_id):
...
+
+            actions = [
+                {"action_type": "add_input", "type": "data", "label": "foo"},
+                {"action_type": "update_step_label", "label": "bar", "step": {"label": "foo"}},
+            ]
+
+        :type dry_run: bool
+        :param dry_run: When true, perform a dry run where the existing
+                        workflow is preserved. A new workflow is
+                        allocated instead.
```
simonbray, line 761:
```suggestion
        :param dry_run: When true, perform a dry run where the existing
                        workflow is preserved. The refactored workflow
                        is returned in the output of the method, but not saved
                        on the Galaxy server.
```

simonbray: I don't think a dry_run creates a new workflow? @rikeshi please correct me if I'm wrong.

Also, could you add `rtype` for this method please?

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,88 @@ def test_run_workflow(self):
...
+        assert invoke_response['state'] == 'new', invoke_response
+
+    def test_create_workflow_from_history(self):
+        path = test_util.get_abspath(os.path.join('data', 'paste_columns.ga'))
+        wf1 = self.gi.workflows.import_workflow_from_local_path(path)
+        history_id = self.gi.histories.create_history(name="test_wf")['id']
+        dataset_id = self._test_dataset(history_id)
+        dataset = {'src': 'hda', 'id': dataset_id}
+
+        _job = sorted(self.gi.jobs.get_jobs(), key=lambda x: x['create_time'], reverse=True)[0]
```
simonbray, line 218: I think this line can be removed?

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,88 @@ def test_run_workflow(self):
...
+            inputs={'Input 1': dataset, 'Input 2': dataset},
+            history_id=history_id,
+            inputs_by='name',
+        )
+
+        time.sleep(5)
+        job = sorted(self.gi.jobs.get_jobs(), key=lambda x: x['create_time'], reverse=True)[0]
+
+        for _ in range(20):
+            # job = sorted(self.gi.jobs.get_jobs(), key=lambda x: x['create_time'], reverse=True)[0]
```
simonbray, line 231: can you remove commented out lines please?

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,88 @@ def test_run_workflow(self):
...
+
+        self.gi.workflows.invoke_workflow(
+            wf1['id'],
+            inputs={'Input 1': dataset, 'Input 2': dataset},
+            history_id=history_id,
+            inputs_by='name',
+        )
+
+        time.sleep(5)
+        job = sorted(self.gi.jobs.get_jobs(), key=lambda x: x['create_time'], reverse=True)[0]
```
simonbray, line 228:
```suggestion
        job = sorted(self.gi.jobs.get_jobs(), key=lambda x: x['create_time'], reverse=True)[0]
        job_id = gi.invocations.show_invocation(inv_id)['steps'][-1]['job_id']
```
Can you do something like this?

bioblend/_tests/GalaxyTestBase.py
simonbray: > Should we maybe move and expose this as a convenience `wait_for_invocation()` method of `InvocationClient` ?

Sounds good to me

bioblend/galaxy/workflows/__init__.py
rikeshi: I was not sure, the code on the Galaxy side did not have it documented. This is what I made of it. You are likely correct.

Yes, I will add it.

bioblend/_tests/TestGalaxyWorkflows.py
rikeshi: Yes, I must have missed it.

bioblend/galaxy/workflows/__init__.py
simonbray: I checked to make sure, it doesn't create a new workflow :+1:

bioblend/_tests/TestGalaxyWorkflows.py
rikeshi: `self.gi.invocations.show_invocation` is only availabe in Galaxy 19.09+ so tests would fail for all older versions.

simonbray: Ah, got it. Then this is an okay solution. It would be nice if get_jobs allowed filtering, but that's not something for this PR.

bioblend/_tests/TestGalaxyWorkflows.py
rikeshi: I removed the check of steps in `wait_for_invocation`.

bioblend/_tests/GalaxyTestBase.py
```diff
@@ -25,3 +26,23 @@ def _test_dataset(self, history_id, contents="1\t2\t3", **kwds):
...
+        galaxy_version = os.environ.get('GALAXY_VERSION', None)
+        is_newer = galaxy_version == 'dev' or galaxy_version >= 'release_19.09'
+        show_invocation = self.gi.invocations.show_invocation if is_newer else self.gi.workflows.show_invocation
+        args = [invocation['id']] if is_newer else [invocation['workflow_id'], invocation['id']]
+        for _ in range(20):
+            invocation = show_invocation(*args)
+            if invocation["state"] == "scheduled":
```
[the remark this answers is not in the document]
rikeshi: Ok, I changed it.

bioblend/_tests/GalaxyTestBase.py
rikeshi: It has been moved to InvocationClient

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,93 @@ def delete_workflow(self, workflow_id):
...
+            ]
+
+        :type dry_run: bool
+        :param dry_run: When true, perform a dry run where the existing
+                        workflow is preserved. The refactored workflow
+                        is returned in the output of the method, but not saved
+                        on the Galaxy server.
+
+        :rtype: dict
+        :return: Dictionary containing the action executions and the refactor workflow.
```
simonbray, line 765:
```suggestion
        :return: Dictionary containing logged messages for the executed actions
                 and the refactored workflow.
```

simonbray (approved):

simonbray:
Looks good to me, thanks @rikeshi!

Any other comments @nsoranzo?

bioblend/galaxy/invocations/__init__.py
```diff
@@ -247,8 +250,37 @@ def get_invocation_biocompute_object(self, invocation_id):
...
+        :type invocation: dict
+        :param invocation: Invocation to wait for.
+
+        :type timeout_seconds: int
+        :param timeout_seconds: Timeout in seconds. If the invocation is not scheduled
+                                after this timeout, an InvocationNotScheduledException
+                                is raised.
+        """
+        galaxy_version = os.environ.get('GALAXY_VERSION', None)
+        is_newer = galaxy_version == 'dev' or galaxy_version >= 'release_19.09'
```
simonbray, line 266: On reading this again, I think this is too complicated, `/api/workflows/.../invocations/...` might be deprecated in newer versions, but it still is totally functional. I would just rely on this so we don't have to check the Galaxy version.

(As an aside, although this way of checking the version might be passing CI tests, normal users don't have the GALAXY_VERSION env variable defined. So this does not work in practice, the correct way would be with `gi.config.get_version()`)

bioblend/galaxy/invocations/__init__.py
```diff
@@ -247,8 +250,33 @@ def get_invocation_biocompute_object(self, invocation_id):
...
+        :param timeout_seconds: Timeout in seconds. If the invocation is not scheduled
+                                after this timeout, an InvocationNotScheduledException
+                                is raised.
+        """
+        for _ in range(timeout_seconds * 2):
+            invocation = self.gi.workflows.show_invocation(invocation['workflow_id'], invocation['id'])
+            if invocation['state'] in INVOCATION_TERMINAL_STATES:
+                break
+            time.sleep(.5)
+        invocation = self.gi.workflows.show_invocation(invocation['workflow_id'], invocation['id'])
```
simonbray, line 270:
```suggestion
```

rikeshi: Woudn't it then raise an exception about waiting for `n` seconds while actually only checking for `n - 0.5` seconds?

simonbray: I see what you mean, let's leave it like this then

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,85 @@ def test_run_workflow(self):
...
+            time.sleep(.5)
+
+        wf1 = self.gi.workflows.show_workflow(wf1['id'])
+        datasets = self.gi.histories.show_history(history_id, contents=True)
+        dataset_hids = [dataset['hid'] for dataset in datasets]
+        job_ids = [job['id']]
+        wf2 = self.gi.workflows.create_workflow_from_history(history_id=history_id, workflow_name='My new workflow!',
+                                                             job_ids=job_ids, dataset_hids=dataset_hids)
+        wf2 = self.gi.workflows.show_workflow(wf2['id'])
+        self.assertEqual(wf2['name'], 'My new workflow!')
```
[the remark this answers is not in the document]
rikeshi: I had issues with `steps` being empty when just calling `create_workflow_from_history()`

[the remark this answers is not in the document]
rikeshi: The first assignment to `wf2` can be ommited

bioblend/galaxy/invocations/__init__.py
```diff
@@ -247,8 +249,36 @@ def get_invocation_biocompute_object(self, invocation_id):
         url = self._make_url(invocation_id) + '/biocompute'
         return self._get(url=url)
 
+    def wait_for_invocation(self, invocation, timeout_seconds=10):
```
nsoranzo, line 252: I think we have removed all calls to this method from tests <19.09, so I'd lean towards having `invocation_id` as argument here and call `self.gi.invocations.show_invocation(invocation_id)` below.
Given that all other methods in this module also supports only >=19.09 I think it's fair.

bioblend/galaxy/invocations/__init__.py
```diff
@@ -247,8 +249,36 @@ def get_invocation_biocompute_object(self, invocation_id):
         url = self._make_url(invocation_id) + '/biocompute'
         return self._get(url=url)
 
+    def wait_for_invocation(self, invocation_id, timeout_seconds=10):
+        """
+        Wait for an invocation to be scheduled.
+
+        :type invocation: str
+        :param invocation: Invocation ID to wait for.
```
nsoranzo, line 257:
```suggestion
        :type invocation_id: str
        :param invocation_id: Invocation ID to wait for.
```

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,94 @@ def delete_workflow(self, workflow_id):
...
+        """
+        payload = {
+            'actions': actions,
+            'dry_run': dry_run,
+        }
+        url = '/'.join((self._make_url(workflow_id), 'refactor'))
+        return self._put(payload=payload, url=url)
+
+    def create_workflow_from_history(self, history_id, workflow_name,
+                                     job_ids=None, dataset_hids=None, dataset_collection_hids=None):
```
nsoranzo, line 776:
```suggestion
    def extract_workflow_from_history(self, history_id, workflow_name,
                                      job_ids=None, dataset_hids=None, dataset_collection_hids=None):
```

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,94 @@ def delete_workflow(self, workflow_id):
...
+        """
+        Create a workflow from a history_id.
```
nsoranzo, line 778:
```suggestion
        Extract a workflow from a history.
```

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,94 @@ def delete_workflow(self, workflow_id):
...
+
+        :type history_id: str
+        :param history_id: Encoded history ID
+
+        :type   workflow_name: str
+        :param  workflow_name: Name of the workflow to create when extracting a workflow from history
```
nsoranzo, line 784:
```suggestion
        :param  workflow_name: Name of the workflow to create
```

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,94 @@ def delete_workflow(self, workflow_id):
...
+
+        :type   job_ids: list
+        :param  job_ids: Optional list of jobs to include when extracting a workflow from history
```
nsoranzo, line 787:
```suggestion
        :param  job_ids: Optional list of job IDs to filter the jobs to extract from the history
```

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,94 @@ def delete_workflow(self, workflow_id):
...
+
+    def show_versions(self, workflow_id):
+        """
+        Get versions for a workflow.
+
+        :type workflow_id: str
+        :param workflow_id: Encoded workflow ID
+
+        :rtype: list of dicts
+        :return: List of versions for this workflow ID
```
nsoranzo, line 817:
```suggestion
        :return: Ordered list of version descriptions for this workflow
```

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,76 @@ def test_run_workflow(self):
             self.gi.workflows.run_workflow(wf['id'], None)
 
     def test_invoke_workflow(self):
+        invocation = self._invoke_workflow()
+        assert invocation['state'] == 'new', invocation
+
+    def test_create_workflow_from_history(self):
```
nsoranzo, line 211:
```suggestion
    def test_extract_workflow_from_history(self):
```

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,76 @@ def test_run_workflow(self):
...
+            if _job['id'] == job['id'] and _job['state'] == 'ok':
+                break
+            time.sleep(.5)
+
+        wf1 = self.gi.workflows.show_workflow(invocation['workflow_id'])
+        datasets = self.gi.histories.show_history(invocation['history_id'], contents=True)
+        dataset_hids = [dataset['hid'] for dataset in datasets]
+        job_ids = [job['id']]
+        new_workflow_name = 'My new workflow!'
+        wf2 = self.gi.workflows.create_workflow_from_history(
```
nsoranzo, line 228:
```suggestion
        wf2 = self.gi.workflows.extract_workflow_from_history(
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -247,8 +249,36 @@ def get_invocation_biocompute_object(self, invocation_id):
...
+            invocation = self.gi.invocations.show_invocation(invocation_id)
+            if invocation['state'] in INVOCATION_TERMINAL_STATES:
+                break
+            time.sleep(.5)
+        invocation = self.gi.invocations.show_invocation(invocation_id)
+        if invocation["state"] != 'scheduled':
+            raise InvocationNotScheduledException(" ".join([
+                f"Invocation with ID {invocation['id']} was not scheduled after {timeout_seconds} seconds.",
+                f"Invocation state: {invocation['state']}."
+            ]))
```
nsoranzo, line 274:
```suggestion
            ]))
        return invocation
```

bioblend/_tests/TestGalaxyWorkflows.py
```diff
@@ -205,15 +205,76 @@ def test_run_workflow(self):
...
+        for _ in range(20):
+            [_job] = [_j for _j in self.gi.jobs.get_jobs() if _j['id'] == job['id']]
+            if _job['id'] == job['id'] and _job['state'] == 'ok':
+                break
+            time.sleep(.5)
+
+        wf1 = self.gi.workflows.show_workflow(invocation['workflow_id'])
+        datasets = self.gi.histories.show_history(invocation['history_id'], contents=True)
+        dataset_hids = [dataset['hid'] for dataset in datasets]
+        job_ids = [job['id']]
```
nsoranzo, line 226: What about something like that (not tested)? It would require annotating this test with `@test_util.skip_unless_galaxy('release_19.09')`:
```suggestion
        invocation = self.gi.invocations.wait_for_invocation(invocation['id'])
        wf1 = self.gi.workflows.show_workflow(invocation['workflow_id'])
        datasets = self.gi.histories.show_history(invocation['history_id'], contents=True)
        dataset_hids = [dataset['hid'] for dataset in datasets]
        job_ids = [step['job_id'] for step in invocation['steps]]
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -247,8 +249,36 @@ def get_invocation_biocompute_object(self, invocation_id):
...
+        """
+        Wait for an invocation to be scheduled.
+
+        :type invocation_id: str
+        :param invocation_id: Invocation ID to wait for.
+
+        :type timeout_seconds: int
+        :param timeout_seconds: Timeout in seconds. If the invocation is not scheduled
+                                after this timeout, an InvocationNotScheduledException
+                                is raised.
```
nsoranzo, line 262:
```suggestion
                                is raised.

        :rtype: dict
        :return: Details of the workflow invocation.
```

bioblend/_tests/TestGalaxyWorkflows.py
rikeshi: Testing locally with `dev, py39` I get `bioblend.ConnectionError: Unexpected HTTP status code: 400: {"err_msg": "Malformed id ( None ) specified, unable to decode", "err_code": 400009, "id": "None"}` with `job_ids = [None, None, 'ebfb8f50c6abde6d']`

simonbray: We only want the last job, as the first two are dataset uploads. @rikeshi can you therefore try replacing the last line with `jobs_ids = invocation['steps'][-1]['job_id']`?

rikeshi: `bioblend.ConnectionError: Unexpected HTTP status code: 500: {"err_msg": "Uncaught exception in exposed API method:", "err_code": 0}`

nsoranzo: I've update the suggestion above by adding a ` if step['job_id']` to the list comprehension that is assigned to `job_ids`. This should work and remain generic.

simonbray:
Seems the test failure here is not related to the PR

nsoranzo:
> Seems the test failure here is not related to the PR

Yep, I wouldn't worry about that. There is still an open comment though.

bioblend/galaxy/workflows/__init__.py
```diff
@@ -731,6 +731,94 @@ def delete_workflow(self, workflow_id):
...
+        :param workflow_id: Encoded workflow ID
+
+        :type actions: list of dicts
+        :param actions: Actions to use for refactoring the workflow. The following
+                        actions are supported: update_step_label, update_step_position,
+                        update_output_label, update_name, update_annotation,
+                        update_license, update_creator, update_report, add_step,
+                        add_input, disconnect, connect, fill_defaults, fill_step_defaults,
+                        extract_input, extract_legacy_parameter,
+                        remove_unlabeled_workflow_outputs,
```
simonbray, line 748:
```suggestion
                        remove_unlabeled_workflow_outputs, upgrade_all_steps,
```

bioblend/_tests/TestGalaxyWorkflows.py
rikeshi: I have applied the suggestion but it would not work without waiting for the job state. I did make the loop a bit cleaner. I am not aware of a better way to do this at the moment.

nsoranzo (approved):
Thanks @rikeshi , I've added 2 more commits adding `JobsClient.wait_for_job()` and aligning ``InvocationClient.wait_for_invocation()` to the other `wait_for` methods.

This is ready to merge from my side, great additions, thank a lot for all the work!

rikeshi:
Okay, great! Thanks a lot for all the help @nsoranzo and @simonbray !

PR #370, opened 2021-02-15
replace param 'key' with header 'x-api-key'

rikeshi:
in GalaxyClient.make_get_request()

As mentioned by @nsoranzo in galaxyproject/galaxy#11375. This also fixes the 'key' error when making an API request.
Not sure if I did it right, but it's a starting point.

### DIFF
--- a/bioblend/galaxyclient.py
+++ b/bioblend/galaxyclient.py
@@ -71,7 +71,7 @@ class GalaxyClient:
         headers['x-api-key'] = self.key
         kwargs.setdefault('verify', self.verify)
         kwargs.setdefault('timeout', self.timeout)
-        r = requests.get(url, **kwargs)
+        r = requests.get(url, headers=headers, **kwargs)
         return r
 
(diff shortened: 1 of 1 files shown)


simonbray (approved):

simonbray:
Actually, I guess this applies to all request types, not just GET. Can you update the other methods as well?

simonbray:
You could simply add it to `json_headers` on line 45: https://github.com/galaxyproject/bioblend/blob/8c20180e5a57bc377d0af1e5d0ebea9cd6248884/bioblend/galaxyclient.py#L45

nsoranzo:
@rikeshi Thanks a bunch for starting this! I've pushed a commit applying to all other REST methods and cleaned it up, can you double check please?

rikeshi:
Thanks for cleaning it up @nsoranzo ! Looks good to me.

nsoranzo (approved):

Issue #374, opened 2021-02-25
Method to cancel a job


simonbray:
i.e. `gi.jobs.cancel_job(job_id)`

PR #375, opened 2021-02-26
Added cancel_job() [Issue-#374]

Mcgregor381:


bioblend/galaxy/jobs/__init__.py
```diff
@@ -191,6 +191,25 @@ def get_metrics(self, job_id):
...
+        :rtype: requests.Response
+        :return: the response object.
+
+        .. note::
+         Calling ``get_state()`` to determine current job's state
+         if the state is not ok then make a delete_request
+        """
+
+        if self.get_state(job_id) != "ok":
+            url = '/'.join((self._make_url(job_id), "/discarded"))
```
simonbray, line 210: This API endpoint doesn't seem to exist as far as I can tell? It should be sufficient to make the DELETE request to `/api/jobs/{id}`

bioblend/galaxy/jobs/__init__.py
```diff
@@ -191,6 +191,25 @@ def get_metrics(self, job_id):
...
+            return self.gi.make_delete_request(url)
```
simonbray, line 211: Can you use the hidden `_delete` method here? Not tested but I think `self._delete(id=job_id)` would be enough.

simonbray (commented):
Thanks @Mcgregor381!

simonbray:
This needs a test as well, i.e. a new `test_cancel_job()` method in https://github.com/galaxyproject/bioblend/blob/main/bioblend/_tests/TestGalaxyJobs.py. Would you have time to do this?

You can also test locally by setting up a local Galaxy instance - let me know if you need a hand with this.

bioblend/galaxy/jobs/__init__.py
simonbray: as documented here: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/jobs.py#L244

bioblend/galaxy/jobs/__init__.py
Mcgregor381: that makes a lot more sense to use _delete, makes this bit code a bit cleaner and easier to read and resolve the issue mentioned in the prior comment about the API endpoint.

Mcgregor381:
I should have time to do it to set up a local instance of the application for testing though once I have time and I'd gladly accept the help in setting an instance up.

as for the unit test, I'm not 100% sure how I would implement one for this.

simonbray:
> I should have time to do it to set up a local instance of the application for testing though once I have time and I'd gladly accept the help in setting an instance up.

Sure, ping me when you are ready and we can have a zoom call

> as for the unit test, I'm not 100% sure how I would implement one for this.

something like:

- run a tool with `gi.tools.run_tool()` to create a job - I would use this tool: https://github.com/galaxyproject/galaxy/blob/dev/test/functional/tools/cat_data_and_sleep.xml, as it will give time to cancel the job while it is still running.
- assert the state is `running`
- run your new method
- assert the state is `discarded`

We can also talk more about this via zoom if needed.

Mcgregor381:
@simonbray Sorry for the slow reply, I busy taking part in a CTF event over the weekend.
as for setting up a zoom call, I'm free most weekdays after 3:30 PM (GTM) and would be happy to work to your schedule 😄

simonbray:
I'll write you an email

simonbray:
@Mcgregor381 I went ahead and finished this off - thanks for the PR!

Mcgregor381:
@simonbray  sorry about not getting it completed uni ended up taking a lot more time than expected 
I did write a test but never end up getting working correctly.  Thank you for finishing this for me

PR #381, opened 2021-03-15
Add function to get /dependency_resolvers/toolbox

cat-bro:
Add file for ToolDependencies API with a function `summarize_toolbox` plus a test file.

https://docs.galaxyproject.org/en/master/api/api.html#module-galaxy.webapps.galaxy.api.tool_dependencies

### DIFF
--- a/bioblend/_tests/TestGalaxyToolDependencies.py
+++ b/bioblend/_tests/TestGalaxyToolDependencies.py
@@ -2,7 +2,7 @@
 """
 Test functions in bioblend.galaxy.tool_dependencies
 """
-from . import GalaxyTestBase
+from . import GalaxyTestBase, test_util
 
 
 class TestGalaxyToolDependencies(GalaxyTestBase.GalaxyTestBase):
--- a/bioblend/_tests/TestGalaxyToolDependencies.py
+++ b/bioblend/_tests/TestGalaxyToolDependencies.py
@@ -13,7 +13,7 @@ class TestGalaxyToolDependencies(GalaxyTestBase.GalaxyTestBase):
         self.assertTrue(isinstance(toolbox_summary, list))
         self.assertGreater(len(toolbox_summary), 0)
 
-        toolbox_summary_by_tool = self.gi.tool_dependencies.summarize_toolbox(index_by='tool')
+        toolbox_summary_by_tool = self.gi.tool_dependencies.summarize_toolbox(index_by='tools')
         self.assertTrue(isinstance(toolbox_summary_by_tool, list))
         self.assertGreater(len(toolbox_summary_by_tool), 0)
         self.assertTrue(isinstance(toolbox_summary_by_tool[0], dict))
@@ -21,7 +21,7 @@ class TestGalaxyToolDependencies(GalaxyTestBase.GalaxyTestBase):
         self.assertTrue(isinstance(toolbox_summary_by_tool[0]['tool_ids'], list))
         tool_id = toolbox_summary_by_tool[0]['tool_ids'][0]
 
-        toolbox_summary_select_tool_ids = self.gi.tool_dependencies.summarize_toolbox(index_by='tool', tool_ids=[tool_id])
+        toolbox_summary_select_tool_ids = self.gi.tool_dependencies.summarize_toolbox(index_by='tools', tool_ids=[tool_id])
         self.assertTrue(isinstance(toolbox_summary_select_tool_ids, list))
         self.assertEqual(len(toolbox_summary_select_tool_ids), 1)
         self.assertEqual(toolbox_summary_select_tool_ids[0]['tool_ids'][0], tool_id)
--- a/bioblend/_tests/TestGalaxyToolDependencies.py
+++ b/bioblend/_tests/TestGalaxyToolDependencies.py
@@ -7,6 +7,7 @@ from . import GalaxyTestBase
 
 class TestGalaxyToolDependencies(GalaxyTestBase.GalaxyTestBase):
 
+    @test_util.skip_unless_galaxy('release_20.01')
     def test_summarize_toolbox(self):
         toolbox_summary = self.gi.tool_dependencies.summarize_toolbox()
         self.assertTrue(isinstance(toolbox_summary, list))
--- a/bioblend/_tests/TestGalaxyToolDependencies.py
+++ b/bioblend/_tests/TestGalaxyToolDependencies.py
@@ -1,0 +1,26 @@
+
+"""
+Test functions in bioblend.galaxy.tool_dependencies
+"""
+from . import GalaxyTestBase
+
+
+class TestGalaxyToolDependencies(GalaxyTestBase.GalaxyTestBase):
+
+    def test_summarize_toolbox(self):
+        toolbox_summary = self.gi.tool_dependencies.summarize_toolbox()
+        self.assertTrue(isinstance(toolbox_summary, list))
+        self.assertGreater(len(toolbox_summary), 0)
+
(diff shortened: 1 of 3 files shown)


bioblend/_tests/TestGalaxyToolDependencies.py
```diff
@@ -0,0 +1,26 @@
+
+"""
+Test functions in bioblend.galaxy.tool_dependencies
+"""
+from . import GalaxyTestBase
+
+
+class TestGalaxyToolDependencies(GalaxyTestBase.GalaxyTestBase):
+
+    def test_summarize_toolbox(self):
```
simonbray, line 10: Judging by the test failures, I assume the API endpoint was not available in Galaxy versions < 20.01? If so, could you please add the `@test_util.skip_unless_galaxy('release_20.01')` decorator?

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+
+        params = {}
+        if index:
+            params['index'] = str(index)
+        if tool_ids:
+            params['tool_ids'] = ','.join(tool_ids)
+        if resolver_type:
+            params['resolver_type'] = resolver_type
+        if include_containers is not None:
```
simonbray, line 74:
```suggestion
        if include_containers:
```
I think there is no difference between None and False here, right?

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+from bioblend.galaxy.client import Client
+
+
+class ToolDependenciesClient(Client):
+
+    def __init__(self, galaxy_instance):
+        self.module = 'dependency_resolvers'
+        super().__init__(galaxy_instance)
+
+    def summarize_toolbox(self, index=None, tool_ids=None, resolver_type=None, include_containers=None, container_type=None, index_by=None):
```
simonbray, line 13: I'm not sure whether these should all default to None, e.g. `include_containers` can default to requirements.

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        GET /api/dependency_resolvers/toolbox
+
+        Summarize requirements across toolbox (for Tool Management grid). This is an experiemental
```
simonbray, line 17:
```suggestion
        Summarize requirements across toolbox (for Tool Management grid). This is an experimental
```

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+        if tool_ids:
+            params['tool_ids'] = ','.join(tool_ids)
+        if resolver_type:
+            params['resolver_type'] = resolver_type
+        if include_containers is not None:
+            params['include_containers'] = str(include_containers)
+        if container_type:
+            params['container_type'] = container_type
+        if index_by:
+            params['index_by'] = index_by
```
simonbray, line 79: Maybe throw an error if it's not one of tools or requirements?

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def summarize_toolbox(self, index=None, tool_ids=None, resolver_type=None, include_containers=None, container_type=None, index_by=None):
+        """
+        GET /api/dependency_resolvers/toolbox
+
+        Summarize requirements across toolbox (for Tool Management grid). This is an experiemental
+        API particularly tied to the GUI - expect breaking changes until this notice is removed.
+
+        :type index: int
+        :param index: index of the dependency resolver
```
simonbray, line 21: Can you clarify a bit more what this means?

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+                          'model_class': 'NullDependency',
+                          'name': 'galaxy_sequence_utils',
+                          'version': '1.1.4'},
+                          {'cacheable': False,
+                          'dependency_type': None,
+                          'exact': True,
+                          'model_class': 'NullDependency',
+                          'name': 'bx-python',
+                          'version': '0.8.6'}],
+              'tool_ids': ['vcf_to_maf_customtrack1']}]
```
simonbray, line 64: Can you add a note here about the supported Galaxy versions and also that this is an admin-only feature?

simonbray (commented):
Thanks @cat-bro - I'm not really familiar with this endpoint, but I wrote a few comments.

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+class ToolDependenciesClient(Client):
+
+    def __init__(self, galaxy_instance):
+        self.module = 'dependency_resolvers'
+        super().__init__(galaxy_instance)
+
+    def summarize_toolbox(self, index=None, tool_ids=None, resolver_type=None, include_containers=None, container_type=None, index_by=None):
+        """
+        GET /api/dependency_resolvers/toolbox
+
```
simonbray, line 16:
```suggestion
```

bioblend/_tests/TestGalaxyToolDependencies.py
cat-bro: Thanks!  I had not considered earlier versions

bioblend/galaxy/tool_dependencies/__init__.py
cat-bro: Thanks @simonbray, will do.

bioblend/galaxy/tool_dependencies/__init__.py
cat-bro: It's the index of the dependency resolver with respect to the config file.  If you use the default config, index 0 is tool shed repositories, 1 is galaxy packages, 2 is conda with the right version etc.  The output is the dependency resolution with respect to the chosen resolver. It seems easier to document this as 'index of the dependency resolver' than try to accurately explain it - I'm not sure if what I just said about it is 100% correct.

cat-bro:
Thanks @simonbray, apologies for leaving this hanging.  I got `index_by` argument wrong, whoops.  Looks like it passed the test with `tool` instead of `tools` purely by accident.  The api will index by 'tools' for any argument other than requirements.https://github.com/galaxyproject/galaxy/blame/dev/lib/galaxy/webapps/galaxy/api/tool_dependencies.py#L224

Thank you for fixing this.

bioblend/galaxy/tool_dependencies/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, galaxy_instance):
+        self.module = 'dependency_resolvers'
+        super().__init__(galaxy_instance)
+
+    def summarize_toolbox(self, index=None, tool_ids=None, resolver_type=None, include_containers=False, container_type=None, index_by='requirements'):
+        """
+        Summarize requirements across toolbox (for Tool Management grid).
+
+        :type index: int
+        :param index: index of the dependency resolver
```
simonbray, line 18:
```suggestion
        :param index: index of the dependency resolver, with respect to the
            ``container_resolvers_conf.xml`` config file.
```

simonbray:
No worries, thanks for the PR!

simonbray (approved):

bioblend/galaxy/tool_dependencies/__init__.py
cat-bro: I think this probably only works for dependency resolvers so I'll make this 'with respect to the dependency resolvers config file', will update now

PR #385, opened 2021-04-02
Fix failing tests

rikeshi:


simonbray (approved):
@rikeshi what was the matter here? It looks like tests are passing on main?

rikeshi:
The linting and tests of my branches started failing locally and on github because of these lines. I assumed it was related to 723a21b04f01c342024190df21cb35e0156320d6.

nsoranzo:
> 
> 
> @rikeshi what was the matter here? It looks like tests are passing on main?

It was due to a new release of https://pypi.org/project/flake8-bugbear/ on 1st April.

PR #386, opened 2021-04-08
Add missing filtering parameters to the get_datasets method

### LINKED ISSUE #373: Filtering for gi.jobs.get_jobs() and gi.datasets.get_datasets()
It would be good to add filtering for these methods, e.g. for history_id.

rikeshi:
See #373

Added a bunch of new filtering parameters. It would be great to get some feedback on which parameters would be useful additions.

The dataset table in Galaxy does not contains `user_id` so I'm assuming that user parameters are not applicable for datasets but left them in for now.

API endpoint: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/datasets.py#L57
Base filters: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/base.py#L928
Dataset filters: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/managers/datasets.py#L700

Based on the links above, the relevant filtering options already supported by Galaxy seem to be:

- `limit`
- `offset`
- `history_id`
- `id`
- `extension`
- `create_time`
- `update_time`
- `name`
- `state`
- `visible`


simonbray:
I think all the ones already supported by Galaxy in that list make sense to add.

I guess you would make a Galaxy PR for `workflow_id, invocation_id, job_id, dataset_collection_id, user_id, date_range_min, date_range_max, user_details`? I'm not sure if this is worthwhile. Maybe for `workflow_id` and `invocation_id`.

Any thoughts @nsoranzo?

rikeshi:
The endpoint seems to use `create_time` and `update_time` with additional `ge`, `le`, `lt` or `gt` instead of `date_range_min` and `date_range_max`.

nsoranzo:
Filtering by `job_id`, `invocation_id` and probably `workflow_id` make sense to me.

bioblend/galaxy/datasets/__init__.py
```diff
@@ -151,18 +226,54 @@ def get_datasets(self, limit=500, offset=0, history_id=None):
...
-        params = {
+        params: Dict[str, Any] = {
             'limit': limit,
             'offset': offset,
         }
         if history_id:
             params['history_id'] = history_id
+
+        q: List[str] = []
+        qv: List[Any] = []
```
simonbray, line 246: Why not just append to `params` like with `history_id`?

rikeshi: It does not work.

The options `limit`, `offset` and `history_id` are defined as parameters of the `index` endpoint method and can therefore be included directly in `params`. 

The other options are expected as `q={filter-option}-{operator}` and `qv={value}` pairs which are then parsed into filters on the server.

rikeshi: Using a `filters` parameter containing all `q`, `qv` options would be cleaner, but also less self-documenting.

```
class Filter:
    def __init__(self, param: str, val: Any, op: str = 'eq'):
        self.q = f'{param}-{op}'
        self.qv = val if op != 'in' else ','.join(val)
```

```
def get_datasets(..., filters: Optional[List[Filter]] = None) -> list:
    ...
    
    if filters:
        params['q'] = [f.q for f in filters] 
        params['qv'] = [f.qv for f in filters]
    
    return self._get(params=params)
```

```
datasets = self.gi.get_datasets(..., filters=[
    Filter('extension', 'txt'),
    Filter('state', ['ok', 'queued'], op='in'),
    ...
])
```

simonbray:
You could also look like it is done here: https://github.com/galaxyproject/bioblend/blob/4d80b68c085117d98b9d23557c33d5b7f7b6e965/bioblend/galaxy/histories/__init__.py#L68

bioblend/galaxy/datasets/__init__.py
```diff
@@ -141,14 +143,81 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
...
+        deleted: bool = False,
+        visible: bool = True,
+        history_id: Optional[str] = None,
+        workflow_id: Optional[str] = None,
+        job_id: Optional[str] = None,
+        dataset_collection_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        create_time: Optional[Dict[str, str]] = None,
+        update_time: Optional[Dict[str, str]] = None,
+        file_size: Optional[Dict[str, int]] = None,
```
simonbray, line 160: @rikeshi I think you need to update some things here?

Nonetheless it's cool how much is supported by the API already. :+1:

bioblend/galaxy/datasets/__init__.py
```diff
@@ -141,14 +143,81 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
...
+        :param update_time_min: Minimum dataset update date (in YYYY-MM-DD format) to
+          filter on.
+
+        :type update_time_max: str
+        :param update_time_max: Maximum dataset update date (in YYYY-MM-DD format) to
+          filter on.
+
+        :type file_size_min: int
+        :param file_size_min: Minimum size of the dataset's data file to filter on,
+          excluding metadata associated with the dataset.
```
simonbray, line 215:
```suggestion
        :param file_size_min: Minimum size (in bytes) of the dataset's data file to
          filter on, excluding metadata associated with the dataset.
```

bioblend/galaxy/datasets/__init__.py
```diff
@@ -141,14 +143,81 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
...
+
+        :type file_size_max: int
+        :param file_size_max: Maximum size of the dataset's data file to filter on,
+          excluding metadata associated with the dataset.
```
simonbray, line 219:
```suggestion
        :param file_size_max: Maximum size (in bytes) of the dataset's data file to
          filter on, excluding metadata associated with the dataset.
```

bioblend/galaxy/datasets/__init__.py
```diff
@@ -157,20 +226,62 @@ def get_datasets(self, limit=500, offset=0, history_id=None):
...
+        :type user_details: bool
+        :param user_details: If ``True`` and the user is an admin, add the user
+          email to each returned dataset dictionary.
 
-        :rtype: list
-        :return: Return a list of dataset dicts.
+        .. note::
+          The following filtering options can only be used with Galaxy ``release_21.05`` or later:
+            workflow_id, invocation_id, job_id, dataset_collection_id, user_id,
+            date_range_min, date_range_max, user_details
```
simonbray, line 236: @rikeshi I think you need to do some updating here?

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -57,15 +60,86 @@ def test_get_datasets(self):
...
+        self.assertEqual(len(datasets), 1)
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, update_time={'lt': ut})
+        self.assertEqual(datasets, [])
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, update_time={'le': ut})
+        self.assertEqual(len(datasets), 1)
+
+    @test_util.skip_unless_galaxy('release_20.05')
+    def test_get_datasets_extension(self):
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id)
+        print(datasets)
```
simonbray, line 107:
```suggestion
```

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -57,15 +60,86 @@ def test_get_datasets(self):
...
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, extension='txt')
+        self.assertEqual(len(datasets), 1)
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, extension='auto')
```
simonbray, line 110:
```suggestion
        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, extension='bam')
```
Not really a big deal but let's use a real extension here.

rikeshi:
I removed all the parameters that are not yet implemented n Galaxy. The PR should be in a somewhat mergable state now.

simonbray:
@nsoranzo related to this PR - what do you think about deprecating the `contents` parameter of `gi.histories.show_history()`, now we have `gi.datasets.get_datasets(history_id=...)`? I did some testing with some of my large EU histories and the latter is on average 2.5x faster. It would be nice to at least point users towards the alternative, `show_history` is/was one of of my most used bioblend methods.

simonbray (approved):

bioblend/galaxy/datasets/__init__.py
```diff
@@ -141,13 +143,29 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
...
+        offset: int = 0,
+        name: Optional[str] = None,
+        extension: Optional[Union[str, List[str]]] = None,
+        state: Optional[Union[str, List[str]]] = None,
+        visible: bool = True,
+        history_id: Optional[str] = None,
+        create_time_min: str = None,
+        create_time_max: str = None,
+        update_time_min: str = None,
+        update_time_max: str = None,
```
nsoranzo, line 158: Do we want to also add `order` ?

nsoranzo:
> @nsoranzo related to this PR - what do you think about deprecating the `contents` parameter of `gi.histories.show_history()`, now we have `gi.datasets.get_datasets(history_id=...)`? I did some testing with some of my large EU histories and the latter is on average 2.5x faster. It would be nice to at least point users towards the alternative, `show_history` is/was one of of my most used BioBlend methods.

I am bit surprised there is such a big performance difference, given both are a single API call from BioBlend and return the same fields by default. Maybe the Galaxy API implementations can be somewhat unified using the more performant code?
On the BioBlend side we can update the `show_history()` docs to mention that `DatasetClient.get_datasets()` offers more filtering and ordering options, without deprecating it (yet?).

bioblend/galaxy/datasets/__init__.py
```diff
@@ -157,20 +177,104 @@ def get_datasets(self, limit=500, offset=0, history_id=None):
...
+        :param update_time_min: Show only datasets last updated after the provided
+          time and date, which should be formatted as ``YYYY-MM-DDTHH-MM-SS``.
+
+        :type update_time_max: str
+        :param update_time_max: Show only datasets last updated before the provided
+          time and date, which should be formatted as ``YYYY-MM-DDTHH-MM-SS``.
+
+        :type order_by: str
+        :param order_by: Attribute for ordering datasets - must be one of
+          ``create_time`` (default), ``update_time``, ``hid``, ``name``.
```
nsoranzo, line 213: The Galaxy docs for this were wrong..., see https://github.com/galaxyproject/galaxy/pull/11933 , so `extension` and `history_id` are missing.
Additionally, "Orders can be stacked as a comma-separated list of values", so I am leaning towards just pass the value through to Galaxy for validation (dropping `order_descending` as well). What do you think?

simonbray: Thanks for spotting that!

> Orders can be stacked as a comma-separated list of values

I did notice that, I just didn't think it was necessary to support multiple ordering attributes in BioBlend. But if you disagree, I don't have any objection to just letting the user pass whatever they like to `order`.

nsoranzo: > > Orders can be stacked as a comma-separated list of values
> 
> I did notice that, I just didn't think it was necessary to support multiple ordering attributes in BioBlend. But if you disagree, I don't have any objection to just letting the user pass whatever they like to `order`.

I bet there's going to be a user asking for the multiple ordering :laughing: . I'd be in favour of letting the user pass their preferred value.

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -57,15 +62,96 @@ def test_get_datasets(self):
...
+        self.assertEqual(len(datasets), 2)
+
+    @test_util.skip_unless_galaxy('release_19.05')
+    def test_get_datasets_limit_offset(self):
         datasets = self.gi.datasets.get_datasets(limit=0)
         self.assertEqual(datasets, [])
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, offset=1)
+        self.assertEqual(len(datasets), 1)
+
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, name='Pasted Entry')
```
nsoranzo, line 78:
```suggestion
    @test_util.skip_unless_galaxy('release_19.05')
    def test_get_datasets_name(self):
        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, name='Pasted Entry')
```

bioblend/galaxy/datasets/__init__.py
```diff
@@ -157,20 +176,99 @@ def get_datasets(self, limit=500, offset=0, history_id=None):
...
+            qv.append(name)
+        if state:
+            op, val = self._param_to_filter(state)
+            q.append(f'state-{op}')
+            qv.append(val)
+        if extension:
+            op, val = self._param_to_filter(extension)
+            q.append(f'extension-{op}')
+            qv.append(val)
+        if visible:
```
nsoranzo, line 242:
```suggestion
        if visible is not None:
```

bioblend/galaxy/datasets/__init__.py
```diff
@@ -141,13 +143,30 @@ def download_dataset(self, dataset_id, file_path=None, use_default_filename=True
...
 
-    def get_datasets(self, limit=500, offset=0, history_id=None):
+    def get_datasets(
+        self,
+        limit: int = 500,
+        offset: int = 0,
+        name: Optional[str] = None,
+        extension: Optional[Union[str, List[str]]] = None,
+        state: Optional[Union[str, List[str]]] = None,
+        visible: bool = True,
```
nsoranzo, line 153:
```suggestion
        visible: Optional[bool] = None,
```

bioblend/galaxy/datasets/__init__.py
```diff
@@ -157,20 +176,99 @@ def get_datasets(self, limit=500, offset=0, history_id=None):
...
+        :param name: Dataset name to filter on.
+
+        :type extension: str or list of str
+        :param extension: Dataset extension (or list of extensions) to filter on.
+
+        :type state: str or list of str
+        :param state: Dataset state (or list of states) to filter on.
+
+        :type visible: bool
+        :param visible: If ``False``, include datasets marked as not ``visible``.
```
nsoranzo, line 189:
```suggestion
        :param visible: Optionally filter datasets by their ``visible`` attribute.
```

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -57,15 +62,96 @@ def test_get_datasets(self):
...
+        self.assertEqual(len(datasets), 2)
+
+    @test_util.skip_unless_galaxy('release_20.05')
+    def test_get_datasets_visible(self):
+        self.gi.histories.update_dataset(history_id=self.history_id, dataset_id=self.dataset_id, visible=False)
         datasets = self.gi.datasets.get_datasets(history_id=self.history_id)
         self.assertEqual(len(datasets), 1)
+        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, visible=False)
+        self.assertEqual(len(datasets), 2)
+        self.gi.histories.update_dataset(history_id=self.history_id, dataset_id=self.dataset_id, visible=True)
```
nsoranzo, line 132:
```suggestion
        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, visible=True)
        self.assertEqual(len(datasets), 2)
        datasets = self.gi.datasets.get_datasets(history_id=self.history_id, visible=False)
        self.assertEqual(len(datasets), 0)
```

bioblend/galaxy/datasets/__init__.py
```diff
@@ -188,6 +192,18 @@ def get_datasets(
         :type visible: bool
         :param visible: Optionally filter datasets by their ``visible`` attribute.
 
+        :type deleted: bool
+        :param deleted: Optionally filter datasets by their ``deleted`` attribute.
+
+        :type purged: str
```
nsoranzo, line 198:
```suggestion
        :type purged: bool
```

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -57,15 +62,124 @@ def test_get_datasets(self):
...
+        self.gi.histories.delete_dataset(self.history_id, self.dataset_id2)
+        deleted_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, deleted=True)
+        self.assertEqual(len(deleted_datasets), 1)
+        purged_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, purged=True)
+        self.assertEqual(purged_datasets, [])
+        self.gi.histories.delete_dataset(self.history_id, self.dataset_id2, purge=True)
+        purged_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, purged=True)
+        self.assertEqual(len(purged_datasets), 1)
+        self.tearDown()
+        self.setUp()
```
nsoranzo, line 158: I doubt that does anything to the history, I will add suggestions for what I meant.
```suggestion
```

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -57,15 +62,124 @@ def test_get_datasets(self):
...
+        cat1_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, tool_id='cat1')
+        self.assertEqual(cat1_datasets, [])
+        upload1_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, tool_id='upload1')
+        self.assertEqual(len(upload1_datasets), 2)
+        self.gi.histories.update_dataset(self.history_id, self.dataset_id2, tags=['test'])
+        tagged_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, tag='test')
+        self.assertEqual(len(tagged_datasets), 1)
+        self.gi.histories.update_dataset(self.history_id, self.dataset_id2, tags=[])
+        tagged_datasets = self.gi.datasets.get_datasets(history_id=self.history_id, tag='test')
+        self.assertEqual(tagged_datasets, [])
```
nsoranzo, line 171:
```suggestion
```

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -10,14 +10,19 @@
...
-        self.dataset_id = self._test_dataset(self.history_id, contents=self.dataset_contents)
+    @classmethod
+    def setUpClass(cls):
+        super().setUp(cls)
+        cls.history_id = cls.gi.histories.create_history(name='TestDataset')['id']
+        cls.dataset_contents = "line 1\nline 2\rline 3\r\nline 4"
+        cls.dataset_id = cls._test_dataset(cls, cls.history_id, contents=cls.dataset_contents)
+        cls.gi.datasets.wait_for_dataset(cls.dataset_id)
+        cls.dataset_id2 = cls._test_dataset(cls, cls.history_id, contents=cls.dataset_contents)
+        cls.gi.datasets.wait_for_dataset(cls.dataset_id2)
```
nsoranzo, line 21:
```suggestion
    def setUp(self):
        super().setUp()
        self.history_id = self.gi.histories.create_history(name='TestDataset')['id']
        self.dataset_contents = "line 1\nline 2\rline 3\r\nline 4"
        self.dataset_id = self._test_dataset(self.history_id, contents=self.dataset_contents)
        self.dataset_id2 = self._test_dataset(self.history_id, contents=self.dataset_contents)
        self.gi.datasets.wait_for_dataset(self.dataset_id)
        self.gi.datasets.wait_for_dataset(self.dataset_id2)
```

bioblend/_tests/TestGalaxyDatasets.py
```diff
@@ -10,14 +10,19 @@
...
 
-    def tearDown(self):
-        self.gi.histories.delete_history(self.history_id, purge=True)
+    @classmethod
+    def tearDownClass(cls):
+        cls.gi.histories.delete_history(cls.history_id, purge=True)
```
nsoranzo, line 25:
```suggestion
    def tearDown(self):
        self.gi.histories.delete_history(self.history_id, purge=True)
```

simonbray:
Thanks @nsoranzo, I also removed creation of the second dataset from `setUp` as it is only needed for the ordering test.

nsoranzo (approved):

PR #392, opened 2021-04-30
Handle HTTPS Cloudman instances better

edwardsnj:
When starting with configuration option use_ssl True, cloudman must use https to communicate with the instance throughout. Some of the initial setup steps assume only http is needed.

### DIFF
--- a/bioblend/cloudman/__init__.py
+++ b/bioblend/cloudman/__init__.py
@@ -468,6 +468,10 @@ class CloudManInstance(GenericVMInstance):
         self.vm_error = ms.get('error', None)
         public_ip = ms.get('public_ip', None)
         # Update url if we don't have it or is different than what we have
+        # After _set_url is called, self.url will have a scheme, so
+        # self.url is always different than public_ip. This has the
+        # potential to change the self.url scheme from https to http if
+        # which are not in launch and don't have access to self.config.
         if not self.url and (public_ip and self.url != public_ip):
             self._set_url(public_ip)
         # See if the cluster has been initialized
--- a/bioblend/cloudman/__init__.py
+++ b/bioblend/cloudman/__init__.py
@@ -397,7 +397,7 @@ class CloudManInstance(GenericVMInstance):
             # Make sure the URL scheme is defined (otherwise requests will not work)
             if not url.lower().startswith('http'):
                 # Check the config for use_ssl to see whether https scheme is required
-                if self.config.kwargs.get('use_ssl',False):
+                if self.config.kwargs.get('use_ssl', False):
                     url = "https://" + url
                 else:
                     url = "http://" + url
@@ -728,7 +728,7 @@ class CloudManInstance(GenericVMInstance):
         if parameters is None:
             parameters = {}
         req_url = '/'.join((self.cloudman_url, 'root', url))
-        # need to set the user "ubuntu" for auth, and indicate verify=False for HTTPS 
+        # need to set the user "ubuntu" for auth, and indicate verify=False for HTTPS
         # connections to self-signed certificates
         r = requests.get(req_url, params=parameters, auth=("ubuntu", self.password), timeout=timeout, verify=False)
         try:
--- a/bioblend/cloudman/__init__.py
+++ b/bioblend/cloudman/__init__.py
@@ -729,8 +729,12 @@ class CloudManInstance(GenericVMInstance):
             parameters = {}
         req_url = '/'.join((self.cloudman_url, 'root', url))
         # need to set the user "ubuntu" for auth, and indicate verify=False for HTTPS
-        # connections to self-signed certificates
-        r = requests.get(req_url, params=parameters, auth=("ubuntu", self.password), timeout=timeout, verify=False)
+        # connections to self-signed certificate, avoid adding verify=False unless use_ssl
+        # config is set
+        extrakwargs = {}
+        if self.config.kwargs.get('use_ssl', False):
+            extragetkwargs = { 'verify': False }
+        r = requests.get(req_url, params=parameters, auth=("ubuntu", self.password), timeout=timeout, **extragetkwargs)
         try:
             json = r.json()
             return json
--- a/bioblend/cloudman/__init__.py
+++ b/bioblend/cloudman/__init__.py
@@ -367,8 +367,12 @@ class CloudManInstance(GenericVMInstance):
         self.config = kwargs.pop('cloudman_config', None)
         if not self.config:
             self.password = password
+            self.verify = kwargs.get("verify",None)
         else:
             self.password = self.config.password
+            if self.config.kwargs.get('use_ssl'):
+                self.verify = kwargs.get("verify",False)
+        self.authuser = kwargs.get("authuser","")
         self._set_url(url)
 
     def __repr__(self):
@@ -397,7 +401,7 @@ class CloudManInstance(GenericVMInstance):
             # Make sure the URL scheme is defined (otherwise requests will not work)
             if not url.lower().startswith('http'):
                 # Check the config for use_ssl to see whether https scheme is required
-                if self.config.kwargs.get('use_ssl', False):
+                if self.config and self.config.kwargs.get('use_ssl', False):
                     url = "https://" + url
                 else:
                     url = "http://" + url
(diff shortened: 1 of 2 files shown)


bioblend/cloudman/launch.py
```diff
@@ -858,7 +859,12 @@ def _checkURL(self, url):
...
-            h = HTTPConnection(p[1])
+            if p.scheme == 'http':
+                h = HTTPConnection(p[1])
+            elif p.scheme == 'https':
+                h = HTTPSConnection(p[1])
+            else:
+                raise RuntimeError("Unexpected url scheme: {}".format(p.scheme))
             h.putrequest('HEAD', p[2])
             h.endheaders()
             r = h.getresponse()
```
nsoranzo, line 870: I am not familiar with this part of BioBlend code, but can this use the higher level `requests` that we use elsewhere?
```python
            r = requests.head(url)
            if r.status_code in (200, 401):
```

edwardsnj: Probably, just trying to make a minimal change least likely to affect others. Perhaps the head method was not available when the original code was written? Seems unlikely, but perhaps...

edwardsnj: OK, did some quick testing. The requests.head strategy hangs (due to retries?) if it gets no response from the url as would be the case as the instance is starting up - I think the original strategy is used to so that the method returns False immediately if the instance is not up yet. I'm going to tweak the patch to continue the current code behavior unless the scheme is "https" so that URLs with no scheme that used to (correctly) use HTTP continue to work. Perhaps this was the reason for the odd (to me) failures in the checks.

bioblend/cloudman/__init__.py
```diff
@@ -724,7 +728,9 @@ def _make_get_request(self, url, parameters=None, timeout=None):
         if parameters is None:
             parameters = {}
         req_url = '/'.join((self.cloudman_url, 'root', url))
-        r = requests.get(req_url, params=parameters, auth=("", self.password), timeout=timeout)
+        # need to set the user "ubuntu" for auth, and indicate verify=False for HTTPS
+        # connections to self-signed certificates
+        r = requests.get(req_url, params=parameters, auth=("ubuntu", self.password), timeout=timeout, verify=False)
```
nsoranzo, line 733: Can you make `verify` also configurable (as `use_ssl`)?

bioblend/cloudman/launch.py
nsoranzo: From what I see with an unreachable URL under Linux, both `h.endheaders()` and `requests.head(url)` hang for a while but fail after several seconds with `[Errno 110] Connection timed out` (a `TimeoutError` and `ConnectionError` exception respectively).

nsoranzo: > Perhaps this was the reason for the odd (to me) failures in the checks.

If you mean the failures in https://github.com/galaxyproject/bioblend/actions/runs/802338623 , these are not connected with your changes (the failed tests were not using Cloudman and some BioBlend tests randomly fail unfortunately).

edwardsnj: > From what I see with an unreachable URL under Linux, both `h.endheaders()` and `requests.head(url)` hang for a while but fail after several seconds with `[Errno 110] Connection timed out` (a `TimeoutError` and `ConnectionError` exception respectively).

Hmm. Not sure why my experience was quite different. Also Linux (CentOS 7.9). With requests.head(url) it didn't return until the instance came up, whereas h.endheaders() was called and returned multiple times before the instance came up. Perhaps it is the difference between an IP address that has an instance, but doesn't yet have the cloudman app running, and an IP address which doesn't exist at all? Not sure. Regardless, in my testing the existing approach worked well, while the requests.head(url) didn't work as well.

edwardsnj: > 
> 
> > Perhaps this was the reason for the odd (to me) failures in the checks.
> 
> If you mean the failures in https://github.com/galaxyproject/bioblend/actions/runs/802338623 , these are not connected with your changes (the failed tests were not using Cloudman and some BioBlend tests randomly fail unfortunately).

Good to know. Thanks...

bioblend/cloudman/__init__.py
edwardsnj: OK, will do. I'm a bit puzzled why I need to add the ubuntu user here - how did this ever work without it? Any clue?

edwardsnj: I notice @nsoranzo adding verify and timeout to get requests elsewhere using data-members... perhaps it would be more appropriate to add data members self.verify, self.authuser (naming?) to CloudManInstance and have these set in the constructor or elsewhere for later use? I presume the Cloudman auth user is a property of the base instance configured in whatever cloud and may need to be configurable (ec2-user, ubuntu, centos, admin).

edwardsnj:
Argh. Unfortunately, CloudManInstance may not have access to the launch configuration depending on how it is constructed. The init method already tries to determine whether it has the config, if not it uses keyword arguments. I've added self.verify and self.authuser for use in _make_get_request. If the config sets use_ssl, the default is verify=False, but the init method's arguments override. The user for cloudman authorization is not (as far as I'm aware) in the configuration, so this is only set by the keyword arguments. These changes are becoming more invasive and probably needs someone with more experience with this code-base to make some decisions about style etc.

bioblend/cloudman/__init__.py
```diff
@@ -365,10 +365,17 @@ def __init__(self, url, password, **kwargs):
         else:
             super().__init__(None, None)
         self.config = kwargs.pop('cloudman_config', None)
+        self.use_ssl = False
         if not self.config:
             self.password = password
+            self.verify = kwargs.get("verify",None)
```
nsoranzo, line 371: I'd prefer to always have `self.verify` either True or False. For the rest of BioBlend we default to True (as in `requests.request()`, for better security), but we could default to False if it makes more sense for Cloudman.

bioblend/cloudman/__init__.py
```diff
@@ -365,10 +365,17 @@ def __init__(self, url, password, **kwargs):
...
+            self.use_ssl = kwargs.get("use_ssl",kwargs.get("verify",None) is not None)
         else:
             self.password = self.config.password
+            if self.config.kwargs.get('use_ssl',False):
+                self.verify = kwargs.get("verify",False)
```
nsoranzo, line 376: Should this be
```suggestion
                self.verify = self.config.kwargs.get("verify", False)
```

edwardsnj: There is also the issue that depending on whether this is an instance coming via the launcher or not, we may not have access to the config to determine whether this is a https url or not. I'm not deigning to introduce new user-configuration parameters (though y'all can if you like). In this case, if the user has asked for ssl in the config, default is False (no verify, needed for self signed certificates), but the init arguments, if present can override. Without the launch config, the default is None. So no, this code is as I intend.

bioblend/cloudman/__init__.py
edwardsnj: The issue is (and again, style and consistency matters, and I don't have enough context to choose wisely there) that you asked for verify to be only added for the get request if it was a https url (for http it is irrelevant). The None state allows us to set verify to True or False or to leave it out.

afgane: The default launch process is to use a self-signed cert (because of DNS requirements) so `False` (vs. `True`) would provide more functional capability out of the box. Otherwise, personally, I don't have a strong preference between `False` or `None` given the above comment.

nsoranzo: > The issue is (and again, style and consistency matters, and I don't have enough context to choose wisely there) that you asked for verify to be only added for the get request if it was a https url (for http it is irrelevant).

I probably expressed myself badly, passing `verify` for HTTP requests is absolutely fine (as you said, it's irrelevant and ignored by `requests`).

bioblend/cloudman/__init__.py
nsoranzo: OK, I see what you mean (I think). I've pushed a commit to allow both `use_ssl` and `verify` to be set in the config, but overridable in `CloudManInstance.__init__()` kwargs. Does it make sense and work for you?

bioblend/cloudman/__init__.py
nsoranzo: I don't know the answer to these questions unfortunately, I co-maintain the Galaxy-part of BioBlend but never really used CloudMan.

afgane: `ubuntu` is the one and only username that worked with CloudMan<2.0 regardless of the cloud so we can just keep that here as the default. The new CM is using `admin` but it's proxied via Keycloak and I'm not really sure if this code will even work with it so if it's working for you @edwardsnj, it's fine to keep it like so until/if a larger update comes along.

edwardsnj:
Mostly I'm trying to get out of the game of monkey patching each release. :-) This ended up touching a bunch of places, and I realized that some of the original changes were not actually doing anything useful. I think this form still does what I need it to do, so I'm happy. Thanks for all of your efforts on this... its much harder to make a solution that works well for all use-cases instead of my specific use-case. :-/

PR #395, opened 2021-05-12
[WIP] Add remap option to rerun_invocation() method

simonbray:


bioblend/galaxy/invocations/__init__.py
```diff
@@ -190,23 +192,36 @@ def rerun_invocation(self, invocation_id: str, inputs_update: Optional[dict] = N
...
+            if remap_failures:
+                raise ValueError(f'remap was set to True, but {remap_failures} out of {len(errored_jobs)} errored jobs could not be remapped.')
+            return rerun_jobs
+
         invocation_details = self.show_invocation(invocation_id)
         workflow_id = invocation_details['workflow_id']
         inputs = invocation_details['inputs']
-        wf_params = invocation_details['input_step_parameters']
+        step_ids_to_indices = {step['workflow_step_id']: step_index for step_index, step in enumerate(invocation_details['steps'])}
+        inputs.update({step_ids_to_indices[param['workflow_step_id']]: param['parameter_value'] for _, param in invocation_details['input_step_parameters'].items()})
```
simonbray, line 220: @nsoranzo I found that the `params` option was not working correctly so I removed it (as params can also be submitted via `inputs`) - does this look like an okay alternative?

nsoranzo: Mmmh, what's not working for you? I notice two issues:
- `invocation_details['input_step_parameters']` seems to be never populated by Galaxy, so `params_update` should actually be just `params` as in `invoke_workflow()`
- `payload = {'inputs': inputs, 'params': wf_params}` should have been `payload = {'inputs': inputs, 'parameters': wf_params}`

Does that help?

simonbray: > invocation_details['input_step_parameters'] seems to be never populated by Galaxy, so params_update should actually be just params as in invoke_workflow()

It does, but only if the input parameter is labelled.

> payload = {'inputs': inputs, 'params': wf_params} should have been payload = {'inputs': inputs, 'parameters': wf_params}

Whoops, maybe that was it - I'll try that, thanks!

simonbray: @nsoranzo it seems that `params` doesn't work with runtime parameters, can you confirm if that's the case?

nsoranzo: Ping @mvdbeek

simonbray: At least I couldn't get it to work with a runtime parameter, and the documentation:
```
        The ``params`` dict should be specified as follows::

          {STEP_ID: PARAM_DICT, ...}

        where PARAM_DICT is::

          {PARAM_NAME: VALUE, ...}
```
doesn't really make sense for a runtime parameter.

simonbray: This seems to make it clear that runtime parameters belong in `inputs`, not `parameters`: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/workflow/run_request.py#L282-L285

PR #399, opened 2021-08-03
Fix error authenticating with email and password

simonbray:


bioblend/galaxyclient.py
```diff
@@ -63,10 +63,8 @@ def __init__(self, url, key=None, email=None, password=None, verify=True, timeou
             self._key = None
             self.email = email
             self.password = password
-        self.json_headers = {
-            'Content-Type': 'application/json',
-            'x-api-key': self.key,
-        }
+        self.json_headers = {'Content-Type': 'application/json'}
```
nsoranzo, line 66: Good catch! Can you please add a comment between these 2 lines to explain the reason why `self.json_headers` need to exist before `self.key` is accessed?

nsoranzo (approved):

mvdbeek:
Any chance we could get a new release soon-ish ? Seeing this in the ephemeris tests currently.

nsoranzo:
> Any chance we could get a new release soon-ish ? Seeing this in the ephemeris tests currently.

At last, v0.17.0 is out with this fix, sorry for the long delay!

PR #402, opened 2021-08-11
Add toolshed tests

simonbray:
Some basic non-comprehensive tests to address https://github.com/galaxyproject/bioblend/issues/397, using the test ToolShed (https://testtoolshed.g2.bx.psu.edu/).

I did not write tests for the `update_repository` and `create_repository` methods of the `ToolShedRepositoryClient`, which make POST requests to the ToolShed and require access to an account / API key. We could create one and put the API key in a GitHub secret @nsoranzo? Not sure if this is desirable though.

Some issues I found so far:

- 4 of the filters for `ts.repositories.repository_revisions()` do not work (removed back in 2016: https://github.com/galaxyproject/galaxy/pull/2249)
-  <del>Setting the `deleted` argument of `ts.categories.get_categories()` to `True` leads to a 500 error.</del> I realized this is expected without administrator access.


simonbray:
The test failures for `test_tool_dependency_install`  are not related, they are due to the older version of conda which was installed in Galaxy versions < 19.05 (see https://github.com/galaxyproject/galaxy/commit/25ea088647b09ee9a212120d907430c11afbdf42). It's not that clear to me why these tests have suddenly started failing though.

simonbray:
> The test failures for test_tool_dependency_install are not related, they are due to the older version of conda which was installed in Galaxy versions < 19.05 (see galaxyproject/galaxy@25ea088). It's not that clear to me why these tests have suddenly started failing though.

Alright, they seem to be passing again now. :)

simonbray:
Hi @nsoranzo, I went ahead and merged a few PRs without waiting for your review, but as this one is a bit larger I wondered if you had any comments?

.github/workflows/test.yml
```diff
@@ -129,3 +129,44 @@ jobs:
...
+        include:
+          - os: ubuntu-latest
+            tox_env: py39
+            galaxy_version: dev
+          - os: ubuntu-latest
+            tox_env: py38
+            galaxy_version: dev
+          - os: ubuntu-latest
+            tox_env: py37
+            galaxy_version: dev
```
nsoranzo, line 149:
```suggestion
        galaxy_version: [dev]
        tox_env:
          - py39
          - py38
          - py37
          - py36
```

.github/workflows/test.yml
```diff
@@ -129,3 +129,44 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      # - name: Cache pip dir
+      #   uses: actions/cache@v2
+      #   with:
+      #     path: ~/.cache/pip
+      #     key: pip-cache-${{ matrix.tox_env }}-${{ matrix.galaxy_version }}
```
nsoranzo, line 156: Can this be uncommented?

.github/workflows/test.yml
```diff
@@ -129,3 +129,44 @@ jobs:
...
+          python-version: ${{ steps.get_bioblend_python_version.outputs.bioblend_python_version }}
+      - name: Install tox
+        run: |
+          python3 -m pip install --upgrade pip setuptools
+          python3 -m pip install 'tox>=1.8.0' 'virtualenv>=20.0.14'
+      - name: Run ToolShed tests
+        run: |
+          export BIOBLEND_TOOLSHED_URL=https://testtoolshed.g2.bx.psu.edu/
+          cd "${{ github.workspace }}"
+          tox -e ${{ matrix.tox_env }}
```
nsoranzo, line 172:
```suggestion
          tox -e ${{ matrix.tox_env }}
```

.github/workflows/test.yml
simonbray, line 156:
```suggestion
      - name: Cache pip dir
        uses: actions/cache@v2
        with:
          path: ~/.cache/pip
          key: pip-cache-${{ matrix.tox_env }}-${{ matrix.galaxy_version }}
```

nsoranzo:
> ```
> >       self.assertGreater(int(samtools_search['total_results']), 20)
> E       AssertionError: 7 not greater than 20
> ```
> 
> I guess these tests will be very unstable. I'm not sure how the `search_repositories` API can go from returning >20 repositories to 7.

Mmmh, I was already wondering if we should run a local ToolShed for the tests so we can run also those creating/modifying repositories, maybe we should try that?

bioblend/_tests/test_util.py
```diff
@@ -23,6 +29,18 @@ def skip_unless_cloudman():
...
+def skip_unless_toolshed():
+    """Decorate tests with this to skip the test if a URL for a ToolShed
+    to run the tests is not provided.
+    """
+    if "BIOBLEND_TOOLSHED_URL" not in os.environ:
+        return unittest.skip(NO_TOOLSHED_MESSAGE)
+    elif requests.get(os.environ["BIOBLEND_TOOLSHED_URL"]).status_code != 200:
+        return unittest.skip(INACCESSIBLE_TOOLSHED_MESSAGE)
+    else:
+        return lambda f: f
```
nsoranzo, line 41:
```suggestion

    return lambda f: f
```

.github/workflows/test.yaml
```diff
@@ -114,3 +114,37 @@ jobs:
...
         run: |
           cat galaxy-${{ matrix.galaxy_version }}/*.log
+  test-toolshed:
+    if: (github.event_name != 'schedule' || github.repository_owner == 'galaxyproject') && git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep 'bioblend/toolshed'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        galaxy_version: [dev]
```
nsoranzo, line 124:
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -114,3 +114,37 @@ jobs:
...
+          - py39
+          - py38
+          - py37
+    steps:
+      - uses: actions/checkout@v2
+      - name: Cache pip dir
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          key: pip-cache-${{ matrix.tox_env }}-${{ matrix.galaxy_version }}
```
nsoranzo, line 135:
```suggestion
          key: pip-cache-${{ matrix.tox_env }}
```

.github/workflows/test.yaml
```diff
@@ -114,3 +114,37 @@ jobs:
...
+    if: (github.event_name != 'schedule' || github.repository_owner == 'galaxyproject') && git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep 'bioblend/toolshed'
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        galaxy_version: [dev]
+        tox_env:
+          - py39
+          - py38
```
nsoranzo, line 127:
```suggestion
          - py310
```

run_bioblend_tests.sh
```diff
@@ -146,6 +146,9 @@ else
 fi
 export BIOBLEND_GALAXY_URL=http://localhost:${GALAXY_PORT}
 
+# Set variable for ToolShed tests
+export BIOBLEND_TOOLSHED_URL=https://testtoolshed.g2.bx.psu.edu/
+
```
nsoranzo, line 151: The CI now uses this script, so if you set `BIOBLEND_TOOLSHED_URL` here, all CI jobs will run the TS tests.

```suggestion
```

.github/workflows/test.yaml
```diff
@@ -114,3 +114,37 @@ jobs:
         if: ${{ failure() }}
         run: |
           cat galaxy-${{ matrix.galaxy_version }}/*.log
+  test-toolshed:
```
nsoranzo, line 117: I think we can simplify the CI setup by dropping all the changes to this file, and instead add:
```
          if [ "${{ matrix.galaxy_version }}" = 'dev' ]; then
              export BIOBLEND_TOOLSHED_URL=https://testtoolshed.g2.bx.psu.edu/
          fi
```

to the `Run tests` step above. In this way the TS tests would run just once per Python version, but together with the other Galaxy tests.

bioblend/_tests/test_util.py
```diff
@@ -23,6 +29,18 @@ def skip_unless_cloudman():
...
 
 
+def skip_unless_toolshed():
+    """Decorate tests with this to skip the test if a URL for a ToolShed
+    to run the tests is not provided.
+    """
+    if "BIOBLEND_TOOLSHED_URL" not in os.environ:
+        return unittest.skip(NO_TOOLSHED_MESSAGE)
+    elif requests.get(os.environ["BIOBLEND_TOOLSHED_URL"]).status_code != 200:
+        return unittest.skip(INACCESSIBLE_TOOLSHED_MESSAGE)
```
nsoranzo, line 39: I think it's better if the tests fail instead of being skipped in this case?
```suggestion
```

.github/workflows/test.yaml
```diff
@@ -9,8 +9,8 @@ concurrency:
   group: test-${{ github.ref }}
   cancel-in-progress: true
 jobs:
-  test:
-    if: github.event_name != 'schedule' || github.repository_owner == 'galaxyproject'
+  test-galaxy:
+    if: (github.event_name != 'schedule' || github.repository_owner == 'galaxyproject') && git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} | grep 'bioblend/galaxy'
```
nsoranzo, line 13: I'd revert this.

simonbray: Yes, you're right, thanks!

bioblend/_tests/TestToolshed.py
```diff
@@ -0,0 +1,88 @@
...
+        self.assertGreater(len(repositories), 5000)
+        self.assertEqual(repositories[0]["model_class"], "Repository")
+
+        # search_repositories
+        samtools_search = self.ts.repositories.search_repositories("samtools", page_size=5)
+        self.assertGreater(int(samtools_search["total_results"]), 20)
+        self.assertEqual(len(samtools_search["hits"]), 5)
+
+        # show_repository
+        bam_to_sam_repo = [n for n in repositories if n["name"] == "bam_to_sam"][0]
```
nsoranzo, line 45:
```suggestion
        bam_to_sam_repo = [r for r in repositories if r["name"] == "bam_to_sam"][0]
```

bioblend/_tests/TestToolshed.py
```diff
@@ -0,0 +1,88 @@
...
+        # show_repository_revision
+        revision_0 = self.ts.repositories.show_repository_revision(revisions[0]["id"])
+        self.assertEqual(revision_0["id"], revisions[0]["id"])
+
+        # repository_revisions with filters
+        for filter in ["downloadable", "malicious", "missing_test_components", "includes_tools"]:
+            filtered_revisions_true = self.ts.repositories.repository_revisions(**{filter: True})
+            filtered_revisions_false = self.ts.repositories.repository_revisions(**{filter: False})
+            self.assertEqual(len(filtered_revisions_false) + len(filtered_revisions_true), len(revisions))
+
```
nsoranzo, line 70: I think we shouldn't encourage the use of the `repositories.repository_revisions()` endpoint, given it can easily return tens of thousand of records, so I'd probably drop this part.

```suggestion
```

nsoranzo:
Thanks Simon!

PR #406, opened 2021-08-31
Add GET /api/tools/{tool_id}/reload

cat-bro:
Add `reload` function to bioblend.galaxy.tools.

This reloads a single tool and can be useful if a wrapper has been altered.  I can't think of a way to test this.

### DIFF
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -105,6 +105,14 @@ class ToolClient(Client):
         url = self._make_url(tool_id) + '/requirements'
         return self._get(url=url)
 
+    def reload(self, tool_id):
+        """
+        Reload specified tool.
+        This functionality is available only to Galaxy admins.
+        """
+        url = self._make_url(tool_id) + '/reload'
+        return self._get(url=url)
+
     def get_citations(self, tool_id: str) -> List[dict]:
         """
         Get BibTeX citations for a given tool ID.
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -105,7 +105,7 @@ class ToolClient(Client):
         url = self._make_url(tool_id) + '/requirements'
         return self._get(url=url)
 
-    def reload(self, tool_id):
+    def reload(self, tool_id: str) -> dict:
         """
         Reload the specified tool in the toolbox. Any changes that have been made to the wrapper
         since the tool was last reloaded will take effect.
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -123,7 +123,7 @@ class ToolClient(Client):
                        'version': '3.4+galaxy1'}}
         """
         url = self._make_url(tool_id) + '/reload'
-        return self._get(url=url)
+        return self._put(url=url)
 
     def get_citations(self, tool_id: str) -> List[dict]:
         """
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -107,7 +107,8 @@ class ToolClient(Client):
 
     def reload(self, tool_id):
         """
-        Reload the specified tool in the toolbox.
+        Reload the specified tool in the toolbox. Any changes that have been made to the wrapper
+        since the tool was last reloaded will take effect.
         This functionality is available only to Galaxy admins.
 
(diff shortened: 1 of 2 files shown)


bioblend/galaxy/tools/__init__.py
```diff
@@ -105,6 +105,14 @@ def requirements(self, tool_id):
         url = self._make_url(tool_id) + '/requirements'
         return self._get(url=url)
 
+    def reload(self, tool_id):
+        """
+        Reload specified tool.
+        This functionality is available only to Galaxy admins.
+        """
+        url = self._make_url(tool_id) + '/reload'
+        return self._get(url=url)
```
simonbray, line 114: This should surely be a PUT request. Would you be happy to make a PR to Galaxy to fix this upstream?

simonbray:
> This reloads a single tool and can be useful if a wrapper has been altered. I can't think of a way to test this.

I guess the best you can do is to reload a tool (even if it does not actually update) and just check that you get the output you are expecting.

cat-bro:
Thanks @simonbray, I had not thought of looking at the response!

bioblend/galaxy/tools/__init__.py
```diff
@@ -105,6 +105,14 @@ def requirements(self, tool_id):
         url = self._make_url(tool_id) + '/requirements'
         return self._get(url=url)
 
+    def reload(self, tool_id):
+        """
+        Reload specified tool.
+        This functionality is available only to Galaxy admins.
+        """
```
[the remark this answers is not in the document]
cat-bro: Thanks! This has been added.

bioblend/galaxy/tools/__init__.py
```diff
@@ -105,6 +105,14 @@ def requirements(self, tool_id):
         url = self._make_url(tool_id) + '/requirements'
         return self._get(url=url)
 
+    def reload(self, tool_id):
+        """
+        Reload specified tool.
```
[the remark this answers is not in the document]
cat-bro: It's not updating the version necessarily.  It's updating the tool to be exactly what is in the wrapper.  Galaxy loads all of the tool wrappers at startup.  If a tool wrapper is changed, it will not be updated in the toolbox unless it is reloaded.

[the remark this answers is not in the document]
simonbray: Could you write one or two sentences to convey that in the docstring?

bioblend/galaxy/tools/__init__.py
mvdbeek: It should be a PUT, and we don't limit the method type, so this should already work, also against old Galaxy instances. ... the issues is that a PUT/POST and DELETE request without parameters in the response body are failing. That's a bug for which a fix would be difficult to backport to all bioblend supported versions and ensure everyone is running with the fix, but this will work:
```suggestion
        return self._put(url=ur, payload={}l)
```

I assume that's why other methods in this class are also setting `payload={}l`

mvdbeek: Actually, it looks like bioblend somehow "smuggles" in a `b"null"` in the request body ... I'm not sure if that is valid or not ... probably not ?. But I guess we don't need to fix that right away given the workaround works.

bioblend/galaxy/tools/__init__.py
```diff
@@ -105,6 +105,26 @@ def requirements(self, tool_id):
...
+        Reload the specified tool in the toolbox. Any changes that have been made to the wrapper
+        since the tool was last reloaded will take effect.
+        This functionality is available only to Galaxy admins.
+
+        :type tool_id: str
+        :param tool_id: id of the requested tool
+
+        :rtype: dict
+        :param: dict containing the id, name, and version of the reloaded tool.
+          For example:
```
nsoranzo, line 119:
```suggestion
          For example::
```

bioblend/galaxy/tools/__init__.py
```diff
@@ -105,6 +105,26 @@ def requirements(self, tool_id):
         url = self._make_url(tool_id) + '/requirements'
         return self._get(url=url)
 
+    def reload(self, tool_id):
```
nsoranzo, line 108:
```suggestion
    def reload(self, tool_id: str) -> dict:
```

cat-bro:
thanks @nsoranzo , @mvdbeek and @simonbray !

nsoranzo (approved):

PR #407, opened 2021-09-24
Only use json.dump if payload is not None

mvdbeek:
Otherwise we send a request body that is `b"null"` (instead of `b""`), which is a problem for the Galaxy API. This means we could drop the `payload={}` we have in many places of the code.

### DIFF
--- a/bioblend/galaxy/client.py
+++ b/bioblend/galaxy/client.py
@@ -154,7 +154,7 @@ class Client:
                 bioblend.log.warning(msg)
                 time.sleep(retry_delay)
 
-    def _post(self, payload, id=None, deleted=False, contents=None, url=None,
+    def _post(self, payload=None, id=None, deleted=False, contents=None, url=None,
               files_attached=False):
         """
         Do a generic POST request, composing the url from the contents of the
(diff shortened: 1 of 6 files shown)


nsoranzo (approved):
Thanks for the fix!

> This means we could drop the payload={} we have in many places of the code.

I think it would be good to drop these as part of this PR. If you don't have the bandwidth, let me know and I can have a look.

mvdbeek:
I will do that

bioblend/galaxy/client.py
```diff
@@ -154,7 +154,7 @@ def _get(self, id=None, deleted=False, contents=None, url=None,
                 bioblend.log.warning(msg)
                 time.sleep(retry_delay)
 
-    def _post(self, payload, id=None, deleted=False, contents=None, url=None,
+    def _post(self, payload=None, id=None, deleted=False, contents=None, url=None,
```
nsoranzo, line 157: For this to work, `make_post_request()` needs to be slightly modified to make its `payload` parameter optional.
Then the tests should pass.

mvdbeek: I've rebased that into the last commit.

nsoranzo (approved):
Thanks @mvdbeek !

Issue #194, opened 2016-08-09
Inconsistent parameter names


afgane:
I just noticed that a number of methods use `toolShed_id` as a parameter (it looks like it's all the methods named `show_repository`; e.g., https://github.com/galaxyproject/bioblend/blob/master/bioblend/galaxy/toolshed/__init__.py#L38), which is (1) inconsistent with the rest of the code; and (2) not PEP8 compliant. Should this be deprecated in favor of `tool_shed_id` or we just forget it?

nsoranzo:
I don't like `toolShed_id` either, but I guess it is part of our API at this point unfortunately.

PR #410, opened 2021-10-15
Small code cleanup

Martmists-GH:
Notes:

- contextlib.suppress was added in 3.4 to avoid try-except-pass statements
- `toolShed_id` was inconsistent with the project and pep8's naming conventions
- yaml.load is considered unsafe, and yaml.safe_load should be used instead for remote data

### DIFF
--- a/bioblend/cloudman/launch.py
+++ b/bioblend/cloudman/launch.py
@@ -1,7 +1,6 @@
 """
 Setup and launch a CloudMan instance.
 """
-import contextlib
 import datetime
 import socket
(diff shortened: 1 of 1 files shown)


martmists-gh:
> Can you please revert the changes about `toolShed_id`? Given it's used as a parameter name, changing it can break user code that use it as keyword argument when calling the methods. I agree it's inconsistent with the project naming conventions, but I prefer to not break our API needlessly.

Ah, I made this change as I saw it in #194, but I'll revert it.

PR #414, opened 2021-12-16
Class method 'build' added to get a tool model including dynamic parameters. 

jaidevjoshi83:
I added the `build` method, which is equivalent to the internal galaxy API functionality `build`. It's useful to get a tool model including dynamic parameters and updated values, repeats block, etc. based on a bioblend API call. A use case can be found in my current project (https://github.com/jaidevjoshi83/galaxylab.git) where bioblend API is being used to create galaxy Tool GUI within Jupyterlab Notebooks.  Since tool forms need to be updated based on selected dynamic parameters therefor this functionality is very useful. 

Thanks

### DIFF
--- a/bioblend/galaxy/tools/__init__.py
+++ b/bioblend/galaxy/tools/__init__.py
@@ -190,36 +190,42 @@ class ToolClient(Client):
         return self._get(id=tool_id, params=params)
 
     def build(self, tool_id, inputs=None, tool_version=None, history_id=None):
-
         """
         This method returns the tool model, which includes an updated input parameter array for the given tool,
         based on user-defined "inputs".
 
         :type inputs: dict
-        :param inputs: (optional) inputs for the payload. Possible values are the
-          default 'legacy' (where inputs nested inside conditionals
-          or repeats are identified with e.g. '<conditional_name>|<input_name>')
-          or '21.01' (where inputs inside conditionals or repeats are nested elements).
+        :param inputs: (optional) inputs for the payload.
           For example::
 
-              {"num_lines": "1",
-              "input": {"values": [{"src": "hda", "id": "4d366c1196c36d18"}]},
-              "seed_source|seed_source_selector": "no_seed",}
+            {
+                "num_lines": "1",
+                "input": {
+                    "values": [
+                        {
+                            "src": "hda",
+                            "id": "4d366c1196c36d18"
+                        }
+                    ]
+                },
+                "seed_source|seed_source_selector": "no_seed",
+            }
 
         :type tool_id: str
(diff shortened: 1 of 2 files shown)


bioblend/galaxy/tools/__init__.py
```diff
@@ -189,6 +189,41 @@ def show_tool(self, tool_id, io_details=False, link_details=False):
...
+
+        if tool_version:
+            params['tool_version'] = tool_version
+
+        if history_id:
+            params['history_id'] = history_id
+
+        url = '/'.join((self.gi.url, 'tools', tool_id, 'build'))
+
+        return self._post(payload=params, url=url)
```
simonbray, line 225:
```suggestion
        return self._get(payload=params, url=url)
```
This should be a GET as far as I can see from https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/api/tools.py#L125?

bioblend/galaxy/tools/__init__.py
```diff
@@ -189,6 +189,41 @@ def show_tool(self, tool_id, io_details=False, link_details=False):
         params['link_details'] = link_details
         return self._get(id=tool_id, params=params)
 
+    def build(self, tool_id, inputs=None, tool_version=None, history_id=None):
+
+        """
+        Get updated tool parameters of a given tool.
```
simonbray, line 195: Could the docstring be extended a bit, to explain what this method should be used for? An example of the `inputs` dict and the output dict would also be nice.

bioblend/galaxy/tools/__init__.py
blankenberg: This is possibly a doc error on the Galaxy end.

The Galaxy UI itself uses POST. I would think there might be issues using GET due to potential limitations in character lengths.

simonbray: It seems the request type is not actually specified: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/webapps/galaxy/buildapp.py#L416. A small test also reveals that POST and GET give the same result.

bioblend/galaxy/tools/__init__.py
jaidevjoshi83: @simonbray I added the descriptions for the method. Also, included detail description, examples and a test for build method. Thanks for reviewing the PR and your valuable suggestions.

jaidevjoshi83: @simonbray Please let me know if you have any other comments or suggestions thank you.

simonbray (approved):

PR #424, opened 2022-06-29
tus upload support

natefoo:
This feels a little clunky and I started with it in the tools module because that's where the rest of the upload methods are, but perhaps this deserves its own module, or belongs in a different existing one?

The yielding method can be used to create a progress bar like [in this example](https://gist.github.com/natefoo/ba1756da6652d6eccbaabfe0519761a3) based on the [resumable_upload.py](https://github.com/galaxyproject/galaxy/blob/6860c57ec573c9f1275b733ff7e60731404a7520/scripts/resumable_upload.py) proof of concept script in Galaxy. We can probably eventually remove resumable_upload.py in favor of wherever this gxupload script ends up.

Do we have server version checks somewhere in bioblend? I did some cursory searching and didn't find any, but it would be awesome to default to have some kind of default upload method that uses tus on 22.01+ and the old method on older Galaxies.

@mvdbeek the metadata supplied with tus is not necessary now that we post to the fetch API after uploading, right?

**TODO:**
- [ ] find the right module
- [x] add additional options from old upload like space_to_tab
- [ ] put gxupload somewhere (ephemeris?)
- [x] figure out the invocation (`--storage`?) to support resuming interrupted uploads.
- [x] document method params

### DIFF
--- a/bioblend/galaxyclient.py
+++ b/bioblend/galaxyclient.py
@@ -9,6 +9,7 @@ import base64
 import contextlib
 import json
 import logging
+from typing import Optional
 from urllib.parse import urljoin
 
 import requests
@@ -24,7 +25,6 @@ from bioblend.util import FileStream
 log = logging.getLogger(__name__)
 
 UPLOAD_CHUNK_SIZE = 10**7
-UPLOAD_ENDPOINT = "/upload/resumable_upload"
 
 
 class GalaxyClient:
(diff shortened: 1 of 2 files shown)


mvdbeek:
That's right, we're not doing anything with it for now.

bioblend/galaxyclient.py
```diff
@@ -256,6 +260,18 @@ def make_patch_request(self, url, payload=None, params=None):
             status_code=r.status_code,
         )
 
+    def get_tus_uploader(self, path, url=None, storage=None, metadata=None, chunk_size=None):
```
nsoranzo, line 263: Is there an use case to have `url` as a parameter instead of hardcoded?

natefoo: Probably not although in theory you could have the tus endpoint be in a non-standard location.

natefoo:
Resuming appears to work if the storage option is used. I have updated the gxupload.py gist with some additional features.

I think this is ready if people are happy with leaving this in the tools module, although a test down both upload paths would be nice. gxupload itself can either go in ephemeris or its own package.

natefoo:
I started writing a test for the yield method and realized maybe it is not that useful since it doesn't return any upload details. So the alternative for things like gxupload that want progress might be to remove the yield method and have the caller get the uploader, perform the upload loop, and then call the post-upload `POST` to the fetch API itself.

natefoo:
Not thrilled with the `post_to_fetch()` name but this is otherwise cleaner (and all the changes are covered by existing tests).

mvdbeek (approved):
Looks great to me, thanks @natefoo!

nsoranzo (approved):

nsoranzo:
@natefoo I've added some type annotations, and made some further consolidation, let me know if that's OK!

natefoo:
Your changes look great, thank you @nsoranzo!

Issue #423, opened 2022-06-29
Drop cloudlaunch support ?


mvdbeek:
@afgane @nuwang is it ok if we drop cloudman support and eliminate the boto dependency ? I think you decomissioned cloudlaunch, right ?

mvdbeek:
(I mean https://github.com/galaxyproject/bioblend/tree/main/bioblend/cloudman, maybe I should have said cloudlaunch ?)

nsoranzo:
@nuwang Is there anything in `bioblend/cloudman/` that is worth keeping? I can do the code removal once we are sure what to drop exactly.

nuwang:
Not any longer, so the whole package can be dropped.

PR #459, opened 2022-08-30
Drop support for CloudMan

nsoranzo:
CloudMan and CloudLaunch are deprecated, see https://galaxyproject.org/blog/2021-10-sunsetting-cloudlaunch/

Fix https://github.com/galaxyproject/bioblend/issues/423 .


nuwang (approved):
Looks good!

PR #460, opened 2022-09-09
Handle delete returning code 202

innovate-invent:
deleting a history returns code 202 in Galaxy 22.05


nsoranzo:
Thanks, this looks fine to me, but I wonder why this doesn't show up in the CI testing. Do you have a traceback?

mvdbeek:
You'll need to enable Celery to defer purging.

nsoranzo:
@mvdbeek So we should merge this, right?

PR #461, opened 2022-10-07
Add argument `require_exact_tool_versions` to `invoke_workflow`

cat-bro:
Add argument `require_exact_tool_versions` (new in 22.05) to invoke_workflow.  This is true by default but can be set to False in the params.

### DIFF
--- a/bioblend/galaxy/workflows/__init__.py
+++ b/bioblend/galaxy/workflows/__init__.py
@@ -293,6 +293,7 @@ class WorkflowClient(Client):
         allow_tool_state_corrections: bool = False,
         inputs_by: Optional[str] = None,
         parameters_normalized: bool = False,
+        require_exact_tool_versions: bool = True,
     ) -> dict:
         """
         Invoke the workflow identified by ``workflow_id``. This will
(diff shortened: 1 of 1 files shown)


bioblend/galaxy/workflows/__init__.py
```diff
@@ -486,6 +492,8 @@ def invoke_workflow(
             payload["allow_tool_state_corrections"] = allow_tool_state_corrections
         if inputs_by is not None:
             payload["inputs_by"] = inputs_by
+        if not require_exact_tool_versions:
+            payload["require_exact_tool_versions"] = require_exact_tool_versions
```
nsoranzo, line 496:
```suggestion
        payload["require_exact_tool_versions"] = require_exact_tool_versions
```

We are not doing it consistently, but I think it's better to pass boolean values in the payload unless there's a good reason not to.

cat-bro: Thanks @nsoranzo, I was in two minds about that

nsoranzo (approved):

nsoranzo:
Thanks @cat-bro !

PR #469, opened 2023-04-27
allow changing extra_user_preferences via update_user call.

uwwint:
I have an extra_user_preference configured, relevant section of user_preferences_extra_conf.yml of my galaxy config:
```yaml
preferences:
    bpadp:
      description: Bpa Data Portal API token
      inputs:
        - name: api_token
          label: Api Token
          type: text
          required: False
```

This translates into a preference key called "bpadp|api_token" which cannot be passed into UserClient.update_user as a kwarg ("|" is a forbidden character in python variables). I updated update_user to accept a dict as second positional parameter, allowing updating such preferences Ala:
`update_user("someID",{"bpadp|api_token":"value"})`

The change is backwards compatible and still supports kwargs passing.

### DIFF
--- a/bioblend/galaxy/users/__init__.py
+++ b/bioblend/galaxy/users/__init__.py
@@ -231,14 +231,16 @@ class UserClient(Client):
         url = self._make_url(user_id) + "/api_key"
         return self._get(url=url)
 
-    def update_user(self, user_id: str, update_data: dict = {}, **kwargs: Any) -> Dict[str, Any]:
+    def update_user(self, user_id: str, user_data: Optional[Dict] = None, **kwargs: Any) -> Dict[str, Any]:
         """
         Update user information. You can either pass a dict of all attributes you want to change, or provide them as kwargs.
         For attributes that cannot be expressed as a kwarg (extra_user_preferences use a | sign) pass them as update_data.
-        Some of the attributes that can be modified are documented below.
 
         :type user_id: str
         :param user_id: encoded user ID
+
+        :type user_data: dict
+        :param user_data: a dict containing the values to be updated, eg. { "username" : "newUsername", "email": "new@email" }
 
         :type username: str
         :param username: Replace user name with the given string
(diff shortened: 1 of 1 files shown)


nuwang (commented):
I'm thinking that the extra parameter should be called `user_preferences` or something, but otherwise, LGTM.

bioblend/galaxy/users/__init__.py
```diff
@@ -248,5 +249,6 @@ def update_user(self, user_id: str, **kwargs: Any) -> Dict[str, Any]:
         :rtype: dict
         :return: details of the updated user
         """
+        update_data.update(kwargs)
```
nsoranzo, line 252:
```suggestion
        if update_data is None:
            update_data = {}
        update_data.update(kwargs)
```

bioblend/galaxy/users/__init__.py
```diff
@@ -231,10 +231,11 @@ def get_or_create_user_apikey(self, user_id: str) -> str:
...
+    def update_user(self, user_id: str, update_data: dict = {}, **kwargs: Any) -> Dict[str, Any]:
         """
-        Update user information. Some of the attributes that can be
-        modified are documented below.
+        Update user information. You can either pass a dict of all attributes you want to change, or provide them as kwargs.
+        For attributes that cannot be expressed as a kwarg (extra_user_preferences use a | sign) pass them as update_data.
+        Some of the attributes that can be modified are documented below.
 
         :type user_id: str
         :param user_id: encoded user ID
```
nsoranzo, line 241: Can you add documentation for `update_data` between `user_id` and `username`?

nsoranzo (commented):
Thanks for the fix!

uwwint:
Hi guys, 

- added documentation
- renamed update_data to user_data
- move from dict = {} to Optional[Dict] = None

KR,
Uwe

nsoranzo:
Thanks for the contribution @uwwint !

Issue #471, opened 2023-06-06
Extend bioblend to support the new notification API 


heisner-tillman:
This makes use of the new `Notification System`, which is not yet merged into the dev branch of galaxy.
For reference look at [https://github.com/galaxyproject/galaxy/pull/15663](url)

List of operations to add:

- [ ] `get_notifications_status`

- [ ] `send_notification`

- [ ] `get_notification_preferences`

- [ ] `update_notification_preferences`

- [ ] `get_user_notifications`

- [ ] `get_broadcasted`

- [ ] `get_all_broadcasted`

- [ ] `show_notification`

- [ ] `update_broadcasted_notification`

- [ ] `update_user_notification`

- [ ] `update_user_notifications`

- [ ] `delete_user_notification`

- [ ] `delete_user_notifications`

- [ ] `broadcast_notification`

nsoranzo:
> The bioblend operations in regards to the Notification-API only work with Galaxy version >= 23.1. Is there a way to make them inaccessible if the Galaxy version is older?

I don't think that's necessary.

> `@test_util.skip_unless_galaxy("release_23.1")` I found this for the tests. Is there something like this for the operations?

You can add a note to the class or method docstrings, see e.g. https://github.com/galaxyproject/bioblend/blob/main/bioblend/galaxy/jobs/__init__.py#L229-L230

heisner-tillman:
> You can add a note to the class or method docstrings

I have done this.

PR #472, opened 2023-06-06
Integrate get_notification_status operation from the Notification API

heisner-tillman:
This is a part of #471 

## Summary

- [x] Add get_notification_status operation
- [x] Add test for get_notification_status operation
- [x] Add NotificationClient to GalaxyClient

## How to test the changes

Run the corresponding test file:
bioblend/_tests/TestGalaxyNotifications.py


bioblend/_tests/TestGalaxyNotifications.py
```diff
@@ -0,0 +1,553 @@
...
+            self.skipTest("This Galaxy instance is not configured to use notifications.")
+
+        # fetch the empty notification status
+        now = datetime.utcnow()
+        status = self.gi.notifications.get_notification_status(since=now)
+
+        # check the datatype
+        assert type(status["total_unread_count"]) == int
+        assert type(status["broadcasts"]) == list
+        assert type(status["notifications"]) == list
```
bernt-matthias, line 101:
```suggestion
        assert isinstance(status["total_unread_count"], int)
        assert isinstance(status["broadcasts"], list)
        assert isinstance(status["notifications"], list)
```

bioblend/galaxy/notifications/__init__.py
```diff
@@ -0,0 +1,568 @@
...
+
+from bioblend.galaxy.client import Client
+
+if TYPE_CHECKING:
+    from bioblend.galaxy import GalaxyInstance
+
+
+class NotificationClient(Client):
+    """
+    This endpoint only works on Galaxy 23.01 or later.
```
bernt-matthias, line 24:
```suggestion
    This endpoint only works on Galaxy 23.1 or later.
```

bioblend/_tests/template_galaxy.ini
```diff
@@ -21,3 +21,4 @@ enable_beta_workflow_modules = True
 master_api_key = $BIOBLEND_GALAXY_MASTER_API_KEY
 enable_quotas = True
 cleanup_job = onsuccess
+enable_notification_system = true
```
bernt-matthias, line 24: Guess the ini is only needed for tests against old Galaxy versions which do not support notifications anyway?

bernt-matthias (commented):
Can you try to rebase? Maybe this solves the test problems.

bioblend/_tests/template_galaxy.ini
heisner-tillman: Makes sense. I removed it

bioblend/_tests/TestGalaxyNotifications.py
heisner-tillman: I adjusted the tests

bioblend/galaxy/notifications/__init__.py
heisner-tillman: I rewrote the docstring

heisner-tillman:
> try to rebase
Have done that.

bernt-matthias:
Tests fail. Guess you need to add `typing-extensions` to https://github.com/galaxyproject/bioblend/blob/fe48906a9f9ce3b38ffbfc7f7418d5d84fec61be/setup.cfg#L69

bioblend/_tests/template_galaxy.ini
heisner-tillman: I also removed it from the galaxy_ini.yml. I guess that was an issue. Also the failed test (test_update_broadcasted_notification) lacked the following condition (which I added now):

```
  if not self.gi.config.get_config()["enable_notification_system"]:
      self.skipTest("This Galaxy instance is not configured to use notifications.")
  if not self.gi.users.get_current_user()["is_admin"]:
      self.skipTest("This tests requires the current user to be an admin, which is not the case.")
```
As such it was not skipped and failed with the following error:
`Unexpected HTTP status code: 403: {"err_msg":"Notifications are disabled in this Galaxy.","err_code":403004}`

I added `enable_notification_system = true` to the template_galaxy.yaml file again. The tests now succeed locally.

I would be grateful if you could explain why the condition is required in the .yaml file but not in the .ini

bernt-matthias: > I would be grateful if you could explain why the condition is required in the .yaml file but not in the .ini

I was just **guessing** that the ini file is only used for older Galaxy versions which do not support yaml config files. .. And these old versions do not support the notification system anyway.

If I'm right it should not change anything when removed from the ini file...

heisner-tillman: It works now. Thanks for the suggestion :)

bgruening:
Green! 🍏

bioblend/_tests/TestGalaxyNotifications.py
```diff
@@ -0,0 +1,557 @@
...
+
+from bioblend.galaxy import GalaxyInstance
+from . import (
+    GalaxyTestBase,
+    test_util,
+)
+
+
+class TestGalaxyNotifications(GalaxyTestBase.GalaxyTestBase):
+    @test_util.skip_unless_galaxy("release_23.1")
```
nsoranzo, line 20: You can move this duplicated annotation from each method to the class.

bioblend/galaxyclient.py
```diff
@@ -254,6 +254,8 @@ def make_put_request(self, url: str, payload: Optional[dict] = None, params: Opt
                     body=r.content,
                     status_code=r.status_code,
                 )
+        elif r.status_code == 204:
+            return {}
```
nsoranzo, line 258: What about returning None?

bioblend/galaxy/notifications/__init__.py
```diff
@@ -0,0 +1,577 @@
+"""
+Contains possible interaction dealing with Galaxy notifications.
+"""
+
+from datetime import datetime
+from typing import (
+    Any,
+    Dict,
+    List,
```
nsoranzo, line 9: `Literal` is part of Python 3.8 (oldest release supported by BioBlend).

```suggestion
    List,
    Literal,
```

bioblend/galaxy/notifications/__init__.py
```diff
@@ -0,0 +1,577 @@
...
+    Optional,
+    TYPE_CHECKING,
+)
+
+from typing_extensions import Literal
+
```
nsoranzo, line 15:
```suggestion
```

setup.cfg
```diff
@@ -70,6 +70,7 @@ install_requires =
     requests>=2.20.0
     requests-toolbelt>=0.5.1,!=0.9.0
     tuspy
+    typing-extensions
```
nsoranzo, line 73:
```suggestion
```

bernt-matthias:
The root of the problem seems to be that from 24.1 notifications are handled by celery and a POST request to /api/notifications returns AsyncTaskResultSummary (or NotificationCreatedResponse)

To make these tests pass I could disable celery (is it possible to disable certain tasks only?) .. but this breaks many other tests.

davelopez:
Instead of disabling Celery, I suggest using the async flow. When sending a notification, we just need to poll on `api/tasks/{task_id}/state` until we get a `SUCCESS` to check when the notification has been dispatched and then you can check if the user received the notification, etc.

PR #473, opened 2023-06-15
Add reload_toolbox method

abretaud:
Hi!
Adding this method that could be handy for a usegalaxy.fr script.
Not sure it's the best place, let me know if it should be somewhere else

### DIFF
--- a/bioblend/galaxy/config/__init__.py
+++ b/bioblend/galaxy/config/__init__.py
@@ -73,7 +73,7 @@ class ConfigClient(Client):
         url = self.gi.url + "/whoami"
         return self._get(url=url)
 
-    def reload_toolbox(self) -> dict:
+    def reload_toolbox(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


bioblend/galaxy/config/__init__.py
```diff
@@ -72,3 +72,13 @@ def whoami(self) -> dict:
...
         return self._get(url=url)
+
+    def reload_toolbox(self) -> None:
+        """
+        Reload the Galaxy toolbox (but not individual tools)
+
+        :rtype: None
+        :return: None
+        """
+        url = self.gi.url + "/toolbox"
```
nsoranzo, line 83: I think this should be:
```suggestion
        url = f"{self._make_url()}/toolbox"
```
?

nsoranzo (approved):
I am disabling the `func-returns-value` mypy error code for tests on the main branch.

Thanks @abretaud !

PR #478, opened 2023-08-21
Allow to keep quota's default value

bernt-matthias:
in the current implementation, it was impossible to update a quota and to leave the `default = "no"`.

was documented correctly but likely undesired .. for instance, it seemed impossible to just change the value of an existing non-default quota.

Edit: you can see it failing over here https://github.com/galaxyproject/bioblend/pull/479. I cherry-picked the test to this PR

TODO: change the default of the default parameter?

### DIFF
--- a/bioblend/_tests/TestGalaxyQuotas.py
+++ b/bioblend/_tests/TestGalaxyQuotas.py
@@ -44,6 +44,31 @@ class TestGalaxyQuotas(GalaxyTestBase.GalaxyTestBase):
         assert quota["operation"] == "-"
         assert quota["description"] == "asdf"
 
+    def test_update_quota_nondefault(self):
+        """
+        try to update a non-default quota (and leave it non-default)
+        """
+        # 1st make it non-default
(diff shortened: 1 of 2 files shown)


bioblend/_tests/TestGalaxyQuotas.py
```diff
@@ -44,6 +44,31 @@ def test_update_quota(self):
...
+            amount="100 GB",
+        )
+        assert f"has been renamed to '{self.quota_name}'" in response
+        assert f"Quota '{self.quota_name}' is no longer the default for registered users." in response
+
+        # update it leaving it non-default
+        response = self.gi.quotas.update_quota(
+            self.quota["id"],
+            name=self.quota_name + "-updated",
+        )
```
bernt-matthias, line 67:
```suggestion
            default = None
       ) 
```

bernt-matthias:
I think this has been superseded by https://github.com/galaxyproject/bioblend/pull/489

PR #482, opened 2023-11-10
Add support for API auth with access token

nuwang:
Depends on: https://github.com/galaxyproject/galaxy/pull/16977

This is the line where the tus client seems to rely on `x-api-key` that I mentioned @mvdbeek: https://github.com/galaxyproject/bioblend/blob/933b910c8192abf3d97b6458c9aa2b7abd989aed/bioblend/galaxyclient.py#L329


bioblend/galaxyclient.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         self,
         url: str,
         key: Optional[str] = None,
+        token: Optional[str] = None,
```
nsoranzo, line 37: Tests are failing because you need to add this parameter also to `bioblend/galaxy/__init__.py:GalaxyInstance.__init__()`, `bioblend/galaxy/objects/galaxy_instance.py:GalaxyInstance.__init__()` and `bioblend/toolshed/__init__.py:ToolShedInstance.__init__()` methods, which were calling this one but passing email to token, password to email and so forth.
Also, can you rebase on the latest `main`? I've just pushed commit 14d9e029c7cfe58f4b46114f4e6a5510fdba23b4 which change these methods to pass the parameters by keyword.
I'd also move `token` after `password` (actually after `*` in the rebase) to not break the order of the positional parameters.

bioblend/toolshed/__init__.py
```diff
@@ -19,6 +19,7 @@ def __init__(
         email: Optional[str] = None,
         password: Optional[str] = None,
         *,
+        token: Optional[str] = None,
```
nsoranzo, line 22: Actually I am not sure that https://github.com/galaxyproject/galaxy/pull/16977 adds support for authenticating via token also to the ToolShed API?

nuwang: Thanks for catching that.

nsoranzo (commented):
I guess something will need to be done also for the tus upload https://github.com/galaxyproject/bioblend/blob/main/bioblend/galaxyclient.py#L330 ?

Ping @natefoo

nsoranzo:
@nuwang Do you want to rebase this now that the Galaxy PR has been merged?

bioblend/galaxyclient.py
```diff
@@ -73,13 +74,18 @@ def __init__(
         # password and grab user's key before first request.
         if key:
             self._key: Optional[str] = key
+        elif token:
+            self.token: Optional[str] = token
```
nsoranzo, line 78:
```suggestion
```

bioblend/galaxyclient.py
```diff
@@ -73,13 +74,18 @@ def __init__(
...
         else:
             self._key = None
             self.email = email
             self.password = password
         self.json_headers: dict = {"Content-Type": "application/json"}
         # json_headers needs to be set before key can be defined, otherwise authentication with email/password causes an error
-        self.json_headers["x-api-key"] = self.key
+        if token:
+            self.json_headers["Authorization"] = f"Bearer {self.token}"
```
nsoranzo, line 86:
```suggestion
            self.json_headers["Authorization"] = f"Bearer {token}"
```

bioblend/galaxyclient.py
nuwang: @nsoranzo Do you recall why you suggested removing these two lines?

nsoranzo: Probably because `self.token` is not (currently) used anywhere.

nsoranzo (approved):

PR #483, opened 2023-12-08
Time filters for histories.get_histories

cat-bro:
Add filtering for `create_time-ge`, `create_time-le`,  `update_time-ge`, `update_time-le` for histories using the same keywords as used in `get_datasets` filtering. A lot of this is reused from `datasets/__init__.py` .


nsoranzo (approved):
Thanks @cat-bro , perfect!

PR #484, opened 2024-02-29
Add view, keys, limit and offset to get_histories

cat-bro:
Adding `keys`, `limit` and `offset` are necessary for updating Galaxy Australia's history management script to use bioblend.

### DIFF
--- a/bioblend/galaxy/histories/__init__.py
+++ b/bioblend/galaxy/histories/__init__.py
@@ -93,7 +93,7 @@ class HistoryClient(Client):
         create_time_max: Optional[str] = None,
         update_time_min: Optional[str] = None,
         update_time_max: Optional[str] = None,
-        view: Optional[str] = "summary",
+        view: Optional[str] = None,
         keys: Optional[List[str]] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
@@ -155,7 +155,7 @@ class HistoryClient(Client):
         create_time_max: Optional[str] = None,
         update_time_min: Optional[str] = None,
         update_time_max: Optional[str] = None,
-        view: Optional[str] = 'summary',
+        view: Optional[str] = None,
         keys: Optional[List[str]] = None,
         limit: Optional[int] = None,
         offset: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


bioblend/galaxy/histories/__init__.py
```diff
@@ -93,6 +93,10 @@ def _get_histories(
         create_time_max: Optional[str] = None,
         update_time_min: Optional[str] = None,
         update_time_max: Optional[str] = None,
+        view: Optional[Literal["summary", "detailed"]] = None,
+        keys: Optional[List[str]] = None,
+        limit: Optional[int] = None,
+        offset: Optional[int] = None,
         all: Optional[bool] = False,
```
nsoranzo, line 100:
```suggestion
```

bioblend/galaxy/histories/__init__.py
```diff
@@ -93,6 +93,10 @@ def _get_histories(
         create_time_max: Optional[str] = None,
         update_time_min: Optional[str] = None,
         update_time_max: Optional[str] = None,
```
nsoranzo, line 95:
```suggestion
        update_time_max: Optional[str] = None,
        all: Optional[bool] = False,
```

bioblend/galaxy/histories/__init__.py
```diff
@@ -143,6 +155,10 @@ def get_histories(
...
+        view: Optional[Literal["summary", "detailed"]] = None,
+        keys: Optional[List[str]] = None,
+        limit: Optional[int] = None,
+        offset: Optional[int] = None,
         all: Optional[bool] = False,
```
nsoranzo, line 162:
```suggestion
```

bioblend/galaxy/histories/__init__.py
```diff
@@ -143,6 +155,10 @@ def get_histories(
         create_time_max: Optional[str] = None,
         update_time_min: Optional[str] = None,
         update_time_max: Optional[str] = None,
```
nsoranzo, line 157:
```suggestion
        update_time_max: Optional[str] = None,
        all: Optional[bool] = False,
```

cat-bro:
Hi @nsoranzo , both the `update_time_max` filter and the `all` option are important for our history queries. `all` means even histories that are not yours, and will work only if the user is admin. We also need to be able to query for which histories have not been updated since a year ago.

bioblend/galaxy/histories/__init__.py
cat-bro: Thanks @nsoranzo !

nsoranzo (approved):

nsoranzo:
Thanks @cat-bro !

PR #488, opened 2024-06-05
Test purging user without prior deletion

bernt-matthias:


bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -104,6 +104,32 @@ def test_update_user(self):
...
+        if self.gi.config.get_config()["use_remote_user"]:
+            self.skipTest("This Galaxy instance is not configured to use local users")
+        new_username = test_util.random_string()
+        new_user = self.gi.users.create_local_user(
+            new_username, f"{new_username}@example.org", test_util.random_string(20)
+        )
+        new_user_id = new_user["id"]
+        updated_username = test_util.random_string()
+        updated_user_email = f"{updated_username}@example.org"
+        self.gi.users.update_user(new_user_id, username=updated_username, email=updated_user_email)
```
nsoranzo, line 122: Updating user is not needed for this test, I believe?

bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -104,6 +104,32 @@ def test_update_user(self):
...
+        updated_user = self.gi.users.show_user(new_user_id)
+        assert updated_user["username"] == updated_username
+        assert updated_user["email"] == updated_user_email
+
+        # delete and purge user after test (if possile)
+        if self.gi.config.get_config()["allow_user_deletion"]:
```
nsoranzo, line 128: Skip test if this is not true? You can move the check to line 115.

bernt-matthias:
Actually wanted to check if purging is possible without deleting the user. It's not. I tried to adapt the test accordingly.

nsoranzo (approved):

PR #490, opened 2024-09-23
Allow libraries function set_permission public lib and dataset

minhtrung1997:
Hi Galaxy
I work with this API methods using this library. Such a wonderful lib By the way, if I want to make the library and dataset public, I think we can adjust like I committed Long to hear if this will make the user feel better

### DIFF
--- a/bioblend/galaxy/libraries/__init__.py
+++ b/bioblend/galaxy/libraries/__init__.py
@@ -748,13 +748,13 @@ class LibraryClient(Client):
         :return: General information about the library
         """
         payload: Dict[str, List[str]] = {}
-        if access_in:
+        if access_in is not None:
             payload["LIBRARY_ACCESS_in"] = access_in
-        if modify_in:
+        if modify_in is not None:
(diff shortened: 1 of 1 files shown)


bioblend/galaxy/libraries/__init__.py
```diff
@@ -748,13 +748,13 @@ def set_library_permissions(
         :return: General information about the library
         """
         payload: Dict[str, List[str]] = {}
-        if access_in:
```
bernt-matthias, line 751: Before this test failed for `access_in` being `None` or `[]`. Now only for `None`. Is this desired?

minhtrung1997: I think [] make things unrestricted and is desired in create public data library.

bernt-matthias: Can you add a test that shows this?

I'm also not entirely sure what you mean with "I want to make the library and dataset public". Has this not been possible before. Again, a test would be highly appreciated.

PR #492, opened 2024-10-14
Add ``quota_source_label`` parameter to ``QuotaClient.create_quota()``

bernt-matthias:
Currently untested, since the test instance would need an object store setup. If easy to setup I could do it (usually testing is a good idea :) )

### DIFF
--- a/bioblend/galaxy/quotas/__init__.py
+++ b/bioblend/galaxy/quotas/__init__.py
@@ -82,6 +82,7 @@ class QuotaClient(Client):
         default: Optional[DefaultQuotaValues] = "no",
         in_users: Optional[list[str]] = None,
         in_groups: Optional[list[str]] = None,
+        quota_source_label: Optional[str] = None,
     ) -> dict[str, Any]:
         """
         Create a new quota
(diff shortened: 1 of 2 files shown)


bioblend/galaxy/quotas/__init__.py
```diff
@@ -175,6 +181,11 @@ def update_quota(
                         non-default quota will throw an error. Passing None is
                         equivalent to not changing the current status.
 
+        :type default: str
+        :param default: Whether or not this is a default quota. Valid values
+                        are "no", "unregistered", "registered" and None. None is
+                        equivalent to "no".
```
nsoranzo, line 187: This should be the documentation for `quota_source_label` not `default`.

bernt-matthias, line 187:
```suggestion
        :type quota_source_label: str
        :param quota_source_label: If set, quota source label to apply this quota operation to. Otherwise, the default quota is used.

```

nsoranzo (approved):

nsoranzo:
I removed the addition of ``quota_source_label`` to `QuotaClient.update_quota()`, as this doesn't seem to be supported upstream.

PR #498, opened 2025-03-07
Add library dataset deletion route

bernt-matthias:
that is not marked as deprecated

### DIFF
--- a/bioblend/_tests/TestGalaxyFolders.py
+++ b/bioblend/_tests/TestGalaxyFolders.py
@@ -70,10 +70,10 @@ class TestGalaxyFolders(GalaxyTestBase.GalaxyTestBase):
         hda_id = self._test_dataset(history["id"])
 
         # Create 2 library datasets into the library folder
-        ldda1 = self.gi.libraries.copy_from_dataset(
+        ld1 = self.gi.libraries.copy_from_dataset(
             library_id=self.library["id"], dataset_id=hda_id, folder_id=self.folder["id"], message="Added HDA"
         )
-        ldda2 = self.gi.libraries.copy_from_dataset(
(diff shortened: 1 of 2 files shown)


nsoranzo:
Can you create a separate `undelete_library_dataset()` method instead, like we have for histories and quotas?
It seems counter-intuitive to invoke `delete_something()` to undelete it.
Also I'd rather keep using the deprecated API route until `purge` is implemented in the "new" one.

Additionally, it looks like the correct name of the parameter is `purge`, not `purged`, so the implementation has always been buggy.

bernt-matthias:
> Can you create a separate undelete_library_dataset() method instead, like we have for histories and quotas? It seems counter-intuitive to invoke delete_something() to undelete it.

Sure. 

> Also I'd rather keep using the deprecated API route until purge is implemented in the "new" one.

Yeah, whatever `purge` means for library datasets. As so often this remains completely unclear to the API user. 

> Additionally, it looks like the correct name of the parameter is purge, not purged, so the implementation has always been buggy.

Good find. Easy to fix.

nsoranzo:
> > Also I'd rather keep using the deprecated API route until purge is implemented in the "new" one.
> 
> Yeah, whatever `purge` means for library datasets. As so often this remains completely unclear to the API user.
> 
> > Additionally, it looks like the correct name of the parameter is purge, not purged, so the implementation has always been buggy.

It should mark the `ld` as purged and fully delete the underlying dataset (`ld.library_dataset_dataset_association.dataset`): https://github.com/galaxyproject/galaxy/blob/9477c7773e0fdf517c411ac8272e3c5e72b07e3d/lib/galaxy/webapps/galaxy/services/library_contents.py#L219-L231

PR #503, opened 2025-04-24
add decode_id method

simonbray:


nsoranzo:
Thanks @simonbray ! It looks like that in Galaxy < 21.09 the whoami API returned a decoded id (i.e. an int), I'll push a patch.

nsoranzo:
> Thanks @simonbray ! It looks like that in Galaxy < 21.09 the whoami API returned a decoded id (i.e. an int), I'll push a patch.

I've pushed a commit that also adds a ``ConfigClient.encode_id()`` method, wrapping the `/api/configuration/encode/{decoded_id}` endpoint added in Galaxy 24.0 , and uses that in the test.

Using a dataset id won't work as this may be the first test to run, when the Galaxy dataset table is still empty.

nsoranzo (approved):

PR #504, opened 2025-05-06
New functions support uploading and exporting invocations archives

Smeds:
A new function has been added to support uploading archived invocations to Galaxy. The get_invocation_biocompute_object function now accepts a payload parameter, allowing the export format of the archive to be specified. Additionally, _wait_for_short_term_storage in invocations/__init__.py has been updated to stream data back.

### DIFF
--- a/bioblend/galaxy/invocations/__init__.py
+++ b/bioblend/galaxy/invocations/__init__.py
@@ -2,12 +2,15 @@
 Contains possible interactions with the Galaxy workflow invocations
 """
 
+import base64
 import logging
 from typing import (
     Any,
     Optional,
     TYPE_CHECKING,
 )
+
+import requests
 
 from bioblend import (
     CHUNK_SIZE,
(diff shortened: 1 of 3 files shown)


bioblend/galaxy/invocations/__init__.py
```diff
@@ -433,18 +474,24 @@ def _wait_for_short_term_storage(
 
         def check_and_get_short_term_storage() -> dict[str, Any]:
```
Smeds, line 475: Need to update return type. Should I set it to `byte`?

nsoranzo (commented):
Can you please rebase on top of commit 305bed6356dbe5a0cfe02553a029ed6bb58372ad , which I've just pushed to improve `Client._get()` type annotation?

Smeds:
@nsoranzo I think I addressed all your comments

bioblend/galaxy/invocations/__init__.py
```diff
@@ -267,6 +270,37 @@ def cancel_invocation(self, invocation_id: str) -> dict[str, Any]:
...
+        :param url: URL for an exported history archive
+
+        """
+        payload: dict[str, Any] = {
+            "history_id": history_id,
+            "model_store_format": model_store_format,
+        }
+        if file_path:
+            with open(file_path, "rb") as reader:
+                payload["store_content_uri"] = "base64://" + base64.b64encode(reader.read()).decode("utf-8")
```
mvdbeek, line 298: That's loading everything into memory, in the client and once more on the backend, and then another time passing it into the celery task 😨. I think we need to fix the Galaxy API first before we can proceed here.

Smeds: True, that would be a problem for many invocations. I have only been playing around with small datasets.

nsoranzo: @Smeds Would you like to split this PR into 2, with one just containing the changes that don't need the Galaxy API fix?

Smeds: I will split it up and make 2 new pull-requests, before I close this one.

Smeds: #505 first pull-request. Does not depend on an API update

Smeds: #506 second pull-request. This one needs an updated API, also modifications to how it forwards the files.

Smeds: @nsoranzo done

nsoranzo:
Superseded by #505 and #506.

PR #505, opened 2025-05-07
Add a function used to fetch invocation archive.

Smeds:


bioblend/galaxy/client.py
```diff
@@ -92,6 +92,7 @@ def _get(
         params: Optional[dict] = None,
         *,
         json: Literal[False],
+        stream: Literal[False],
```
nsoranzo, line 95:
```suggestion
```

bioblend/galaxy/client.py
```diff
@@ -92,6 +92,7 @@ def _get(
         params: Optional[dict] = None,
```
nsoranzo, line 92:
```suggestion
        params: Optional[dict] = None,
        stream: bool = False,
```

bioblend/galaxy/invocations/__init__.py
```diff
@@ -425,19 +427,51 @@ def _wait_for_short_term_storage(
...
+        def check_and_get_short_term_storage() -> Any:
             if self._get(url=is_ready_url):
-                return self._get(url=url)
+                return self._get(url=url, json=json, stream=stream)
             raise NotReady(f"Storage request {storage_request_id} is not ready")
 
         return wait_on(check_and_get_short_term_storage, maxwait=maxwait, interval=interval)
 
+    def get_invocation_archive(
+        self, invocation_id: str, payload: dict[str, Any], maxwait: float = 1200
```
nsoranzo, line 444: Can you please replace payload with the parameters you want to expose, and then construct the payload inside the method?

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -94,6 +95,25 @@ def test_get_invocation_report(self):
...
+        }
+        self.gi.invocations.wait_for_invocation(invocation["id"])
+        with tempfile.TemporaryDirectory() as folder:
+            file = f"{folder}/temp.rocrate.zip"
+            response = self.gi.invocations.get_invocation_archive(invocation["id"], payload=payload)
+            with open(file, "bw") as archive:
+                for chunk in response.iter_content(chunk_size=8192):
+                    archive.write(chunk)
+            # Verify file is not empty
+            assert os.path.getsize(file) > 0, "Downloaded archive is empty"
```
nsoranzo, line 115: Maybe you can check if it's a Zip file with https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile ?

Smeds: good suggestion

bioblend/galaxy/invocations/__init__.py
Smeds: done

Smeds:
@nsoranzo I have update according to you review comments

nsoranzo:
Thanks @Smeds !

PR #506, opened 2025-05-07
function used to import an invocation archive. 

Smeds:
Needs API updates before it can be merged.


Smeds:
For this to work, we need the following galaxy pull-request [20359](https://github.com/galaxyproject/galaxy/pull/20359) to be merged.

Smeds:
@nsoranzo should I add some comment to the documentation that this function requires version XXX of galaxy?

bioblend/_tests/TestGalaxyInvocations.py
```diff
@@ -112,6 +112,18 @@ def test_get_invocation_jobs_summary(self):
         assert len(step_jobs_summary) == 1
         assert step_jobs_summary[0]["populated_state"] == "ok"
 
+    @test_util.skip_unless_galaxy("release_25.01")
```
nsoranzo, line 115:
```suggestion
    @test_util.skip_unless_galaxy("release_25.1")
```

nsoranzo: Sure, just noting that's the format we are currently using for releases.

PR #534, opened 2026-04-09
Add credential handling to UserClient and ToolClient

dannon:
## Summary

- Add `get_credentials`, `create_credentials`, `select_credential_group`, and `get_credentials_for_tool` methods to `UserClient` wrapping Galaxy's `/api/users/{id}/credentials` endpoints
- Add `credentials_context` parameter to `ToolClient.run_tool()` for passing stored credentials when executing tools that require them
- `get_credentials_for_tool` is a convenience method that builds the `credentials_context` list from stored credentials, ready to pass directly to `run_tool()`

## Test plan

- [ ] Integration tests added for all credential methods (gated on `release_25.1`)
- [ ] Verify `create_credentials` + `select_credential_group` + `get_credentials_for_tool` + `run_tool(credentials_context=...)` workflow against a Galaxy instance with credential support
- [ ] `ruff check` and `ruff format` pass

### DIFF
--- a/bioblend/_tests/TestGalaxyUsers.py
+++ b/bioblend/_tests/TestGalaxyUsers.py
@@ -197,7 +197,7 @@ class TestGalaxyUsers(GalaxyTestBase.GalaxyTestBase):
     def test_get_credentials_for_tool(self):
         user_id = self.gi.users.get_current_user()["id"]
         tool_id = f"test_credential_tool_run_{test_util.random_string()}"
-        self.gi.users.create_credentials(
+        group = self.gi.users.create_credentials(
             user_id=user_id,
             source_type="tool",
             source_id=tool_id,
@@ -206,6 +206,16 @@ class TestGalaxyUsers(GalaxyTestBase.GalaxyTestBase):
             service_version="2.0",
             group_name="default",
             secrets=[{"name": "token", "value": "abc123"}],
+        )
+        # Set the active credential group (not auto-set on creation)
+        creds = self.gi.users.get_credentials(user_id, source_id=tool_id)
+        self.gi.users.select_credential_group(
+            user_id=user_id,
+            source_type="tool",
+            source_id=tool_id,
+            source_version="1.0",
+            user_credentials_id=creds[0]["id"],
+            group_id=group["id"],
         )
         context = self.gi.users.get_credentials_for_tool(user_id, tool_id)
         assert context is not None
@@ -216,3 +226,9 @@ class TestGalaxyUsers(GalaxyTestBase.GalaxyTestBase):
         assert entry["version"] == "2.0"
         assert "id" in entry["selected_group"]
         assert entry["selected_group"]["name"] == "default"
+
+    @test_util.skip_unless_galaxy("release_25.0")
+    def test_get_credentials_for_tool_none(self):
+        user_id = self.gi.users.get_current_user()["id"]
+        context = self.gi.users.get_credentials_for_tool(user_id, "nonexistent_tool_id")
+        assert context is None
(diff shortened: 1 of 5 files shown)


bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -156,3 +156,89 @@ def test_create_user_apikey(self):
...
+                source_version=tool["version"],
+                service_name="test_service",
+                service_version="1.0",
+                group_name="default",
+                variables=[{"name": "api_url", "value": "https://example.org"}],
+                secrets=[{"name": "api_key", "value": "secret123"}],
+            )
+        except ConnectionError as e:
+            if "does not require any credentials" in str(e) or "not defined" in str(e):
+                self.skipTest("Test tool does not have credential definitions")
```
nsoranzo: This and the following tests are always skipped because `random_lines1` doesn't require credentials, can we use `secret_tool` instead (which is used in Galaxy's `test/integration/test_credentials.py`)?

bioblend/galaxy/users/__init__.py
```diff
@@ -257,3 +257,199 @@ def update_user(self, user_id: str, user_data: dict | None = None, **kwargs: Any
...
+        user_id: str,
+        source_type: str,
+        source_id: str,
+        source_version: str,
+        user_credentials_id: str,
+        group_id: str | None,
+    ) -> None:
+        """
+        Select the active credential group for a set of user credentials.
+        This must be called after ``create_credentials()`` before the
```
nsoranzo:
```suggestion
        This must be called after credentials have been created but before the
```

bioblend/_tests/TestGalaxyUsers.py
dannon: Good catch -- switched to secret_tool in 9ea2978. The tests now actually run end-to-end (create + select group + verify) using secret_tool's service1/v1 credential definition, matching the pattern in Galaxy's test_credentials.py.

dannon: Thank you!  I think I've sorted it out, looks like they ran this time and passed.

bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -156,3 +156,80 @@ def test_create_user_apikey(self):
...
+        assert creds == []
+
+    @test_util.skip_unless_galaxy("release_25.1")
+    @test_util.skip_unless_tool("secret_tool")
+    def test_create_and_get_credentials(self):
+        user_id = self.gi.users.get_current_user()["id"]
+        tool_id = "secret_tool"
+        tool = self.gi.tools.show_tool(tool_id)
+        group_name = test_util.random_string()
+        cred = self.gi.users.create_credentials(
```
nsoranzo:
```suggestion
        group = self.gi.users.create_credentials(
```

bioblend/_tests/TestGalaxyUsers.py
```diff
@@ -156,3 +156,80 @@ def test_create_user_apikey(self):
...
+            source_type="tool",
+            source_id=tool_id,
+            source_version=tool["version"],
+            service_name="service1",
+            service_version="v1",
+            group_name=group_name,
+            variables=[{"name": "server", "value": "https://example.org"}],
+            secrets=[{"name": "username", "value": "alice"}, {"name": "password", "value": "s3cret"}],
+        )
+        assert cred["name"] == group_name
```
nsoranzo:
```suggestion
        assert group["name"] == group_name
```

nsoranzo (approved):

bioblend/_tests/TestGalaxyUsers.py
dannon: Thanks for the rename suggestion -- good call for clarity.

nsoranzo:
Thanks @dannon !

dannon:
@nsoranzo Thank you for the careful review and suggestions!

PR #535, opened 2026-04-15
Add UnprivilegedToolsClient and tool_uuid support on run_tool

dannon:
## Summary

- New ``UnprivilegedToolsClient`` wrapping ``/api/unprivileged_tools`` (create/list/show/deactivate) so callers like galaxy-mcp can stop hand-rolling URLs and JSON extraction.
- ``ToolClient.run_tool`` now accepts ``tool_uuid`` and ``tool_version`` kwargs so unprivileged user tools can be run by UUID -- both ``tool_id`` and ``tool_uuid`` go through ``POST /api/tools`` in Galaxy.
- Exactly one of ``tool_id`` or ``tool_uuid`` must be provided; existing positional callers are unaffected.

## Test plan

- [x] ``tox -e lint`` (ruff, flake8, black, isort, mypy) clean
- [ ] ``pytest tests/TestGalaxyUnprivilegedTools.py`` against a Galaxy with unprivileged_tools enabled and the user granted ``USER_TOOL_EXECUTE``
- [ ] Confirm existing ``TestGalaxyTools.py`` still passes (no regression in ``run_tool``)

### DIFF
--- a/bioblend/galaxy/unprivileged_tools/__init__.py
+++ b/bioblend/galaxy/unprivileged_tools/__init__.py
@@ -88,8 +88,8 @@ class UnprivilegedToolsClient(Client):
         """
         Run an unprivileged tool by its ``tool_uuid``.
 
-        Unprivileged tool runs are submitted to the jobs API rather than the
-        regular tools API; the returned shape mirrors
+        Submitted to ``POST /api/tools``, which accepts ``tool_uuid`` as an
+        alternative to ``tool_id``; the returned shape mirrors
         :meth:`bioblend.galaxy.tools.ToolClient.run_tool`.
 
         :type history_id: str
@@ -128,5 +128,5 @@ class UnprivilegedToolsClient(Client):
         if tool_version is not None:
             payload["tool_version"] = tool_version
 
-        url = "/".join((self.gi.url, "jobs"))
+        url = "/".join((self.gi.url, "tools"))
         return self._post(payload, url=url)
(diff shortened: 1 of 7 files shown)


qchiujunhao (commented):
One non-blocking documentation note.

bioblend/galaxy/unprivileged_tools/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, galaxy_instance: "GalaxyInstance") -> None:
+        super().__init__(galaxy_instance)
+
+    def get_user_tools(self, active: bool = True) -> list[dict[str, Any]]:
+        """
+        List the current user's unprivileged tools.
+
+        :type active: bool
+        :param active: If ``True`` (the default), return only active tools;
+          if ``False``, return only deactivated tools.
```
qchiujunhao, line 32: Non-blocking documentation note: on current Galaxy versions, `active=False` does not return tools deactivated through this API. Galaxy filters both `DynamicTool.active` and `UserDynamicToolAssociation.active`, while DELETE only deactivates the user association. Such tools are therefore absent from both `active=True` and `active=False` results and may still resolve by UUID through `/api/tools`. Could we avoid promising “return only deactivated tools” and mention this server-side limitation? Upstream context: https://github.com/galaxyproject/galaxy/pull/22625

qchiujunhao (approved):
lgtm! Thank you! No blocking issues found; the inline documentation note is non-blocking.

